PR #20, opened 2015-01-09
Bindings finished

pickle27:
@BlakeMesdag pointed me to a problem with how I had fixed the initial dirty state for Forms in Shopify and we realized that at the root of the issue was that we didn't have anything to hook onto when the binding for a node was completed (aka the node and its children are bound).

In this PR @BlakeMesdag and I add this functionality `bindingFinished`. What we do is when binding we check if the context has a `bindingFinished` attribute, if it does we pop it off and store it temporarily (this way child nodes don't re-call this). When all the children are bound and we return to the top level where we popped off the function we call it and then re-attach it to the node (this is for re-bindings so it can be re-bound the same way).

We should probably also add a note about this to the readme.

for review
@qq99 @pushrax 
cc @BlakeMesdag

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ $(document).ajaxComplete ->
 
 ## Twine.register
 
-Optionally register a callback that will be called when the context that registers it is finished binding.
+Registers a function to be called when the currently binding node and its children have finished binding.
 
 Example:
 
@@ -106,6 +106,9 @@ Example:
     constructor: ->
       Twine.register ->
         console.log("done")
+
+    # other methods needed in the context
+    # ...
 ```
 
 ```html
--- a/README.md
+++ b/README.md
@@ -95,6 +95,23 @@ $(document).ajaxComplete ->
   Twine.refresh()
 ```
 
+## bindingsFinished
+
+When a node and its children are finished binding Twine checks the context for a function `bindingsFinished` and calls it if its present. This function will only be called once per context.
+
+Example:
+
+```html
+<div define="bindingsFinished: function() { console.log("done"); }"></div>
+
+<script>
+  var foo = { bindingsFinished: function() { console.log("done"); } }
+</script>
+
+<div context="foo">
+</div>
+```
+
 ## Contributing
 
 1. Clone the repo: `git clone git@github.com:Shopify/twine`
--- a/README.md
+++ b/README.md
@@ -95,21 +95,21 @@ $(document).ajaxComplete ->
   Twine.refresh()
 ```
 
-## bindingsFinished
+## Twine.register
 
-When a node and its children are finished binding Twine checks the context for a function `bindingsFinished` and calls it if its present. This function will only be called once per context.
+Optionally register a callback that will be called when the context that registers it is finished binding.
 
 Example:
 
+```coffee
+  class Foo
+    constructor: ->
+      Twine.register ->
+        console.log("done")
+```
+
 ```html
-<div define="bindingsFinished: function() { console.log("done"); }"></div>
-
-<script>
-  var foo = { bindingsFinished: function() { console.log("done"); } }
-</script>
-
-<div context="foo">
-</div>
+<div context='bar' define='{bar: new Foo}'></div>
 ```
 
 ## Contributing
--- a/dist/twine.js
+++ b/dist/twine.js
@@ -61,7 +61,7 @@ Twine.register = function(callback) {
 };
 
 bind = function(context, node, forceSaveContext) {
-  var binding, callback, callbacks, childNode, definition, element, fn, keypath, newContextKey, type, _i, _j, _len, _len1, _ref, _ref1;
+  var binding, callback, callbacks, childNode, definition, element, fn, keypath, newContextKey, type, _i, _j, _len, _len1, _ref, _ref1, _ref2;
   currentBindingCallbacks = [];
   if (node.bindingId) {
     Twine.unbind(node);
@@ -101,8 +101,9 @@ bind = function(context, node, forceSaveContext) {
     bind(context, childNode);
   }
   Twine.count = nodeCount;
-  for (_j = 0, _len1 = callbacks.length; _j < _len1; _j++) {
-    callback = callbacks[_j];
+  _ref2 = callbacks || [];
+  for (_j = 0, _len1 = _ref2.length; _j < _len1; _j++) {
+    callback = _ref2[_j];
     callback();
   }
   currentBindingCallbacks = null;
--- a/dist/twine.js
+++ b/dist/twine.js
@@ -51,7 +51,7 @@ Twine.bind = function(node, context) {
 };
 
 bind = function(context, node, forceSaveContext) {
-  var binding, childNode, definition, element, fn, keypath, newContextKey, type, _i, _len, _ref, _ref1;
+  var binding, childNode, definition, element, finalization, fn, keypath, newContextKey, type, _i, _len, _ref, _ref1;
   if (node.bindingId) {
     Twine.unbind(node);
   }
@@ -79,6 +79,11 @@ bind = function(context, node, forceSaveContext) {
     }
     context = getValue(context, keypath) || setValue(context, keypath, {});
   }
+  finalization = null;
+  if ((context != null ? context.bindingsFinished : void 0) != null) {
+    finalization = context.bindingsFinished;
+    context.bindingsFinished = null;
+  }
   if (element || newContextKey || forceSaveContext) {
     (element != null ? element : element = {}).childContext = context;
     elements[node.bindingId != null ? node.bindingId : node.bindingId = ++nodeCount] = element;
@@ -89,6 +94,10 @@ bind = function(context, node, forceSaveContext) {
     bind(context, childNode);
   }
   Twine.count = nodeCount;
+  if (finalization) {
+    context.bindingsFinished = finalization;
+    finalization.bind(context)();
+  }
   return Twine;
 };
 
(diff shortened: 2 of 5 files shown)


qq99:
Is there any possibility that two+ `Twine.refreshImmediately` calls happen in quick succession, re-evaluating things twice and firing this twice vs. just once?

BlakeMesdag:
Looks like `Twine.refreshImmediately` is synchronous, this code is also synchronous so shouldn't be an issue.

DrewMartin:
So this can only be used by classes that are used as a context? Not the best IMO...

Can we just have things that care call something like `afterBinding` with a callback, and then all callbacks are run after binding is done

BlakeMesdag:
You can arbitrarily define it as well, such as
```html
<div define="bindingsFinished: function() { console.log("done"); }"></div>
```

DrewMartin:
It would still have to be defined as a context though, wouldn't it? Like what you put wouldn't work, you'd have to do:
```html
<div define="bindingsFinished: function() { console.log("done"); }" context="bindingsFinished"></div>
```

pickle27:
nope it can be bare, @DrewMartin check the diff for the tests

qq99:
Can you also add docs on usage to README?

DrewMartin:
I still think it's not great to have a method be automatically called based on its name. I think it would be better to have an instance register itself for callback in the constructor. Kind of like how we do internally for Bindings.onRefresh

BlakeMesdag:
That's another option, this is essentially that without all the callback registration dancing you'd have to do.

lib/assets/javascripts/twine.js.coffee
```diff
@@ -59,6 +64,11 @@ bind = (context, node, forceSaveContext) ->
   # As a result, Twine are unsupported within DocumentFragment and SVGElement nodes.
   bind(context, childNode) for childNode in (node.children || [])
   Twine.count = nodeCount
+
+  if finalization
+    context.bindingsFinished = finalization
+    finalization.bind(context)()
```
pushrax: Why bind it and not just do `finalization.call(context)`?

pushrax:
What @DrewMartin said, or do it the same way event handlers are done, e.g. `<div bindings-complete="do_stuff()"`

Storing it in the context with `define` means that you can accidentally overwrite it from another node.

pickle27:
The examples we gave with the define maybe weren't the best - the reason we did it this way is so that we can use the pattern we have in shopify where we create an object and then use the object name as the context for some nested html and javascript. We want to fire this event on Shopify.Form when the bindings are finished and we didn't want to add a `bindings-complete` everywhere we have defined Shopify.Form. This way we can add a single new method to this object that will get fired when the bindings are done.

We could accomplish this with a DOM event as well but I don't think `bindings-complete` is the right way to do this. 

@pushrax can you clarify your second point about storing the context in the define?

lib/assets/javascripts/twine.js.coffee
pickle27: yup we can deffs do this, we were debating which way to go about it

BlakeMesdag:
@pushrax we pop the callback off the context when we parse the bindings. It's not currently possible to accidentally overwrite it.

pickle27:
When we arrive on the final solution here we should make sure we announce it. Some teams have put some *hacks* in place to do essentially this so it would be cleaning up some technical debt to use a single solution tucked inside of Twine

qq99:
:+1: for an interface like `<div bindings-complete="do_stuff()"...`

BlakeMesdag:
@qq99 but you can just define the function on each node as it currently is

qq99:
Even though it's not possible to accidentally overwrite it, I like that it's got a special keyword and doesn't appear in the `define="{}"` with the rest of your functions for some reason, maybe let's you see it easier in a slew of defines?

Really can go either way though

pickle27:
I really don't like the idea of `<div bindings-complete="do_stuff()"...` because it means we need to add more boiler plate to ever definition of Shopify.Form

qq99:
what's the markup look like with current implementation?  `define="{form: new Shopify.Form(), bindingsComplete: ...}"` ?

pickle27:
@qq99 we just define a method called `bindingsFinished` in Shopify.Form - there is no added markup

qq99:
oh that's pretty sweet, so any object can define that?
That sounds pretty legit then

pickle27:
yup any object can define it.

the `<div define=...` example is just the simplest way to show how this works but its leading everyone astray - I expect 99% of the usage of this will be defining a bindingsFinished on an object that will be used as a context.

qq99:
If we go with this API, then can you also add examples reflecting how we anticipate to use it too?  I can support this format since adding a `bindingsComplete` to various classes seems quite nice for things that need to do this, but it's not obvious from a docs perspective

pickle27:
I have an example of adding to an object in the readme and our tests show both usages - maybe we should remove the example of doing it with the define?

It was really just in the PR comments that we only have the define cause it was shorter to type out for @BlakeMesdag

qq99:
The README says:

```
<script>
 var foo = { bindingsFinished: function() { console.log("done"); } }
</script>

<div context="foo">
</div>
```

but isn't your usage more like:
```
<script>
 var foo = { bindingsFinished: function() { console.log("done"); } }
</script>

<div context="{bar: foo}">
</div>
```
?

pickle27:
my understanding is that it looks for an object with the name of the context but I haven't looked into the code. Have a look at what we did over on the Shopify side: https://github.com/Shopify/shopify/pull/36123

pickle27:
@qq99 is on board now, @pushrax what are your thoughts on the latest discussion here?

DrewMartin:
My biggest issue is that it's not consistent with how Twine is used elsewhere. The only ways to interact with it right now are to have things defined in the DOM (like bind-show) or to explicitly call a method in Twine, like bind and refresh. This adds a third way for one case, when using one of the previous ways would work just as well. Admittedly, having it in the DOM would lead to more places needing to write it, but having it as a function on Twine would require no extra boilerplate for the case you wrote this for.

BlakeMesdag:
Your last point is pretty conflicting, you're saying having a callback stack for each node would require less boilerplate, but you'd need to write those callbacks onto every DOM node you wanted them. That's a lot of boilerplate in itself, its just shifted into the public space (which usually doesn't care).

DrewMartin:
No, you're getting confused between the two alternate suggestions:
* Adding a new binding used for this, which would require adding a bunch of boilerplate, but is more consistent with how Twine handles things
* Adding a function like `Twine.finalize` that takes a callback, which calls it after binding finishes. This would not require more boilerplate, and is still more consistent than the current approach since there are things like reset, bind and refresh that are defined like this.

Either way, I think consistency is more important than having to write some extra boilerplate. Extra boilerplate would only be painful once when first introducing this change, but after that it would be clear what's happening. Or use a callback and it's clear from reading the code.

Having a function called automatically based on its name is too magic-y IMO since this is the only place we'd be doing that. It starts down the path of having too many different ways to do things and makes it less predictable to use.

README.md
```diff
@@ -95,6 +95,28 @@ $(document).ajaxComplete ->
...
 ```
 
+## Twine.register
+
+Optionally register a callback that will be called when the context that registers it is finished binding.
+
+Example:
+
+```html
+<div define="Twine.register(function() { console.log("done"); })"></div>
```
DrewMartin: This is a bad example. Define should return an object. Even if this is returning an object it's probably going to not do what you expect

README.md
```diff
@@ -95,6 +95,28 @@ $(document).ajaxComplete ->
...
+```
+
+```coffee
+  class Foo
+    constructor: ->
+      Twine.register(->
```
DrewMartin: Eww get rid of the parentheses

README.md
```diff
@@ -95,6 +95,28 @@ $(document).ajaxComplete ->
...
+        console.log("done")
+      )
+```
+
+```html
+<div context='bar' define='bar: new Foo'></div>
```
DrewMartin: Thing won't work. It won't crash, but it returns new Foo, not `{bar: new Foo}`

dist/twine.js
```diff
@@ -1,5 +1,5 @@
...
-var attribute, bind, elements, eventName, fireCustomChangeEvent, getContext, getValue, isKeypath, keypathForKey, keypathRegex, nodeCount, preventDefaultForEvent, refreshElement, refreshQueued, rootContext, rootNode, setValue, setupAttributeBinding, setupEventBinding, stringifyNodeAttributes, valueAttributeForNode, wrapFunctionString, _i, _j, _len, _len1, _ref, _ref1,
+var attribute, bind, currentBindingCallacks, elements, eventName, fireCustomChangeEvent, getContext, getValue, isKeypath, keypathForKey, keypathRegex, nodeCount, preventDefaultForEvent, refreshElement, refreshQueued, rootContext, rootNode, setValue, setupAttributeBinding, setupEventBinding, stringifyNodeAttributes, valueAttributeForNode, wrapFunctionString, _i, _j, _len, _len1, _ref, _ref1,
```
DrewMartin: callacks? Callbacks?

README.md
qq99: this line should be `define={bar: new Foo()}` right?

pickle27:
@DrewMartin came and finished pairing with @BlakeMesdag and me. Final look someone?

DrewMartin:
I like it.

yolo ship

pickle27:
ping!

qq99:
LGTM

lib/assets/javascripts/twine.js.coffee
```diff
@@ -32,7 +34,14 @@ Twine.reset = (newContext, node = document.documentElement) ->
...
   bind(context, node, true)
 
+Twine.register = (callback) ->
+  if currentBindingCallbacks
+    currentBindingCallbacks.push(callback)
+  else
+    callback()
+
 bind = (context, node, forceSaveContext) ->
+  currentBindingCallbacks = []
```
pushrax: Would prefer if this was instantiated lazily since it will be rarely used.

README.md
```diff
@@ -95,6 +95,23 @@ $(document).ajaxComplete ->
   Twine.refresh()
 ```
 
+## Twine.register
+
+Optionally register a callback that will be called when the context that registers it is finished binding.
```
pushrax: This is kinda confusing. How about something closer to

> Registers a function to be called when the currently binding node and its children have finished binding.

pushrax:
Looks good for the Shopify.Form use-case. I am not a fan of this though:

```html
<div context='bar' define='{bar: new Foo}'></div>
```

I do understand that it's just an example, and Foo would actually do something other than attach the callback, but we should have something stating you wouldn't actually write this.

In fact, are we currently (ab)using `define` to do this kind of thing? Do we need

```html
<div initialize='do.something.when.binding()'>
```

?

DrewMartin:
I've had the same thought. Being able to execute arbitrary code would be
useful, instead of needing to return an object for define to consume. I
always found alternatives and didn't propose the change to twine though.

pickle27:
@pushrax made your readme change. I added a little note to suggest that the object should do more than just add a callback - what are your guy's thoughts?

pickle27:
are we good to do a gem release?

BlakeMesdag:
Maybe, minor version bump would probably be warranted here.

qq99:
:+1:

pickle27:
I commited a version bump and pushed a tag - can someone release the gem?

qq99:
In the future make sure to bump these too
https://github.com/Shopify/twine/commit/20dd0ba9cc19e59f43cbf7035d7d0bc0736abbbc
and then `bundle install` and commit that too

Gem is up, can you write release notes? 
https://github.com/Shopify/twine/releases/new (input tag and title of `v0.0.11`)

pickle27:
will do and thanks

PR #21, opened 2015-01-20
Guard against null while getting context

qq99:
cc @patrickdonovan 
Fairly self-explanatory, the line above can set `node` to `null` and we don't protect against this, causing the method to throw


pushrax:
Can this be merged?

qq99:
:ship:? cc @Shopify/tnt

DrewMartin:
:+1:

maartenvg:
Is the version number change till cool? If so: :+1:

qq99:
yeah, I had to re-write the version bump

PR #24, opened 2015-02-10
check attribute change should fire  event

celsodantas:
`bind-check` should fire a change event (`bindings:change`) whenever a `check` happen in the node.

Currently this is not happening, so if you have a node like:

```html
<input bind-checked="publishable.isAlreadyPublished()" type="checkbox">
```
Changing the return value of `publishable.isAlreadyPublished()` will check/uncheck the input but no change event is fired.

I need this to fix a bug in Shopify where changing the checkbox using the binding system is not making the form dirty as it does not get any event from the checkbox node.

@qq99 @nsimmons @pushrax @richardmonette

### DIFF
--- a/lib/assets/javascripts/twine.js.coffee
+++ b/lib/assets/javascripts/twine.js.coffee
@@ -286,6 +286,8 @@ setupAttributeBinding = (attributeName, bindingName) ->
       return if newValue == lastValue
       node[attributeName] = lastValue = newValue
 
+      fireCustomChangeEvent(node) if attributeName == 'checked'
+
 for attribute in ['placeholder', 'checked', 'disabled', 'href', 'title', 'readOnly', 'src']
   setupAttributeBinding(attribute, attribute)
 
--- a/dist/twine.js
+++ b/dist/twine.js
@@ -286,7 +286,7 @@ stringifyNodeAttributes = function(node) {
   result = "";
   while (i < nAttributes) {
     attr = node.attributes.item(i);
-    result += "" + attr.nodeName + "='" + attr.textContent + "'";
+    result += attr.nodeName + "='" + attr.textContent + "'";
     i += 1;
   }
   return result;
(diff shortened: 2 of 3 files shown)


lib/assets/javascripts/twine.js.coffee
```diff
@@ -286,6 +287,8 @@ setupAttributeBinding = (attributeName, bindingName) ->
       return if newValue == lastValue
       node[attributeName] = lastValue = newValue
 
+      fireCustomChangeEvent(node) if fireEvent
```
richardmonette: can this not just be if attributeName == 'checked' and then remove line 279?

qq99: :+1: 
Did the change I suggested in Slack work?

nsimmons:
:shipit:

lib/assets/javascripts/twine.js.coffee
```diff
@@ -276,6 +276,7 @@ Twine.bindingTypes =
 
 setupAttributeBinding = (attributeName, bindingName) ->
   booleanAttribute = attributeName in ['checked', 'disabled', 'readOnly']
+  fireEvent = true if attributeName == 'checked'
```
pushrax: To avoid this being `undefined`, you can write `fireEvent = (attributeName == 'checked')`

lib/assets/javascripts/twine.js.coffee
pushrax: oh, +1 to this as well

pushrax:
:+1: to this functionality

lib/assets/javascripts/twine.js.coffee
celsodantas: > Did the change I suggested in Slack work?

nope =(

celsodantas:
done. I'll merge once :green_heart:

PR #25, opened 2015-02-12
version bump to 0.0.14

celsodantas:
@qq99 @richardmonette for sanity check

##  Changes

 - Checkbox toggle now fires `bindings:change` event when a binding causes it to toggle.


qq99:
also do `make .all` and `make .uglify` to update the twine.js and twine.min.js

celsodantas:
done and done

qq99:
:ship:

PR #26, opened 2015-03-02
Add mouseenter mouseleave

tmlayton:
Review @qq99 @pushrax

### DIFF
--- a/lib/assets/javascripts/twine.js.coffee
+++ b/lib/assets/javascripts/twine.js.coffee
@@ -312,5 +312,5 @@ setupEventBinding = (eventName) ->
     return teardown: ->
       $(node).off eventName, onEventHandler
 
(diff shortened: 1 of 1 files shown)


qq99:
:+1:

pushrax:
:+1: assuming we actually need these events for something that mouseover/mouseout will not work for.

qq99:
I was curious what the difference was, researched to find this:

>the mouseover and mouseout events fire for a given element if the user mouses over or out of an element nested within the given element

So it seems like `mouseenter` and `mouseleave` don't fire if you mouse into a child~

:+1: sounds like a legit addition

tmlayton:
It's legit, although I could remove mouseover and mouseout as they were being added for what I will now being using mouseenter and mouseleave for

tmlayton:
`mouseover` and `mouseout` are a bit like spray and pray

qq99:
I'd leave `mosueover` and `mouseout` in

tmlayton:
Will do, okay to merge?

qq99:
:ship:

PR #31, opened 2015-04-08
Speed up attribute loop in bind

DrewMartin:
I was looking for ways to make bind more efficient and this loop stuck out to me. We check all binding types available on twine against all attributes on the node. I tried doing the opposite: looping through the attributes, comparing them against available binding types. It was about 30% faster when binding a large subtree.

jsperf comparison: http://jsperf.com/getting-element-attributes
the few/some/many in the test descriptions refer to the number of attributes on the node being bound. From the results it's apparent that using getAttribute in the list doesn't scale very well. Internally getAttribute probably does a linear search of the attributes. Using the attribute list is both faster overall, and scales better with the number of attributes.

@qq99 @pushrax @richardmonette @celsodantas

### DIFF
--- a/lib/assets/javascripts/twine.js.coffee
+++ b/lib/assets/javascripts/twine.js.coffee
@@ -45,12 +45,15 @@ bind = (context, node, forceSaveContext) ->
   if node.bindingId
     Twine.unbind(node)
 
-  for type, binding of Twine.bindingTypes when definition = node.getAttribute(type)
-    element = {bindings: []} unless element  # Defer allocation to prevent GC pressure
-    fn = binding(node, context, definition, element)
-    element.bindings.push(fn) if fn
+  for attribute in node.attributes
+    if binding = Twine.bindingTypes[attribute.name.toLowerCase()]
+      element = {bindings: []} unless element  # Defer allocation to prevent GC pressure
+      fn = binding(node, context, attribute.value, element)
+      element.bindings.push(fn) if fn
+    else if attribute.name == 'context'
+      newContextKey = attribute.value
 
-  if newContextKey = node.getAttribute('context')
+  if newContextKey
     keypath = keypathForKey(newContextKey)
     if keypath[0] == '$root'
       context = rootContext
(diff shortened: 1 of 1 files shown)


lib/assets/javascripts/twine.js.coffee
```diff
@@ -297,7 +300,7 @@ setupAttributeBinding = (attributeName, bindingName) ->
 
       fireCustomChangeEvent(node) if attributeName == 'checked'
 
-for attribute in ['placeholder', 'checked', 'disabled', 'href', 'title', 'readOnly', 'src']
+for attribute in ['placeholder', 'checked', 'disabled', 'href', 'title', 'readonly', 'src']
```
DrewMartin: Needed for case insensitivity of the bindings

qq99:
screen shot 2015-04-08 at 1 22 53 pm

nice

lib/assets/javascripts/twine.js.coffee
```diff
@@ -293,11 +296,12 @@ setupAttributeBinding = (attributeName, bindingName) ->
       newValue = fn.call(node, context, rootContext)
       newValue = !!newValue if booleanAttribute
       return if newValue == lastValue
+      attributeName = 'readOnly' if attributeName == 'readonly'
```
DrewMartin: This also because `readOnly` is a special attribute on elements (and `readonly` is not equivalent) but the binds should be case-insensitive

pushrax:
:+1: 

Could you post a simple overall perf comparison on a Shopify page?

DrewMartin:
Testing a few of the same pages, average of 5 runs:
(products/1504 has 100 variants)

Page | node.attributes | node.getAttribute() | % of original
--- | --- | --- | ---
home | 15.2 ms | 21.6 ms | 70.4%
products | 49.6 ms | 103.8 ms | 47.8%
products/inventory | 165.4 ms | 243.0 ms | 68.1%
products/1504 | 997.8 ms | 1152.4 ms | 86.6%
products/1504/variants/3056 | 42.4 ms | 52.2 ms |  81.2%
autcomplete w/ 1496 suggestions | 232.4 ms | 338.2 ms | 68.7%

DrewMartin:
CC @Shopify/tnt

pushrax:
Awesome :+1:

qq99:
Great work @DrewMartin

qq99:
What metric do those `ms` represent?

DrewMartin:
That's one call to Twine.bind on initial page visit (except the autocomplete, it's when the bind happens, which is when it is opened)

nsimmons:
:metal: fucking badass.

https://unicorn.shopify.com/posts/15180

DrewMartin:
:ship: ?

maartenvg:
Yes! :sheep: it

PR #32, opened 2015-04-13
Add an init binding

DrewMartin:
Adds an `init` binding. It's similar to `define` except its return value is ignored. One place this would be useful is building an array with bindings, since you can't insert something in an array with define unless you do a hack like `define="array.push(value), {}"`. Simpler to do `init="array.push(value)"`

Some places where this would be useful:
https://github.com/Shopify/shopify/blob/71d34193b66dacefd76f857eab79a8dfcb708ce0/app/views/admin/navigation/_next_nav.html.erb#L364
https://github.com/Shopify/shopify/blob/71d34193b66dacefd76f857eab79a8dfcb708ce0/app/views/admin/home/_welcome_carousel.html.erb#L1
https://github.com/Shopify/shopify/blob/71d34193b66dacefd76f857eab79a8dfcb708ce0/app/views/admin/quotes/_add_shipping.html.erb#L20
https://github.com/Shopify/shopify/blob/71d34193b66dacefd76f857eab79a8dfcb708ce0/app/views/admin/settings/checkout_settings/_checkout.html.erb#L2

@Shopify/tnt

### DIFF
--- a/lib/assets/javascripts/twine.js.coffee
+++ b/lib/assets/javascripts/twine.js.coffee
@@ -286,6 +286,11 @@ Twine.bindingTypes =
     context[key] = value for key, value of object
     return
 
+  eval: (node, context, definition) ->
+    fn = wrapFunctionString(definition, '$context,$root', node)
+    fn.call(node, context, rootContext)
+    return
+
 setupAttributeBinding = (attributeName, bindingName) ->
   booleanAttribute = attributeName in ['checked', 'disabled', 'readOnly']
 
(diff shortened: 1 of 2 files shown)


qq99:
I like it, but would prefer it were called `eval` instead of `init`

celsodantas:
sheep and :+1: to call it `eval` as it's more scary and might prevent people from over using it. =)

DrewMartin:
Renamed to eval

PR #34, opened 2015-04-27
Fals-y values remove the attribute in bind-attribute

DrewMartin:
Right now if an attribute is bound to a fals-y value (other than null or undefined) then it will have that attribute with the value rendered as a string. For example:
```html
<div bind-attribute="{attr: false}"></div>
```
will result in 
```html
<div attr="false"></div>
```

This is a bit awkward in places where we only care about the presence of an attribute, not the actual value, because ideally we'd use a boolean statement in the logic, but if the boolean statement is false we get something like the above. Instead, any false-y value will get replaced with a null, which causes the attribute to be removed.

@richardmonette 
@Shopify/tnt

### DIFF
--- a/lib/assets/javascripts/twine.js.coffee
+++ b/lib/assets/javascripts/twine.js.coffee
@@ -274,7 +274,7 @@ Twine.bindingTypes =
     return refresh: ->
       newValue = fn.call(node, context, rootContext)
       for key, value of newValue when lastValue[key] != value
-        $(node).attr(key, value)
+        $(node).attr(key, value || null)
       lastValue = newValue
 
   define: (node, context, definition) ->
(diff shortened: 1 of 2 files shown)


qq99:
LGTM

maartenvg:
Makes sense to me. The only usage I've seen is where we don't care about false, so we're catching this early enough to make this change to the API :+1:

DrewMartin:
Yeah, it's only used (as far as I'm aware) for tg-remote-noserialize, and doing `<logical statement> || null` is kind of crappy, especially if you want the negation in the statement

pushrax:
I like this approach. :+1: 

> especially if you want the negation in the statement

What do you mean by this?

DrewMartin:
I mean like `<div bind-attribute="{attr: !attrIsDisabled()}">`

pushrax:
Could add a conditional for this: if the value is true set it to an empty string. However, an empty string is falsy, so maybe it makes sense to just use the string "true"

DrewMartin:
yeah setting the value to "true" is fine. It's the setting it to "false" as opposed to unsetting it that's annoying

DrewMartin:
Lol never merged

PR #37, opened 2015-06-17
remove Gemfile.lock

EiNSTeiN-:
Gemfile.lock should not be committed for a gem.

@qq99 for review


qq99:
:+1:

rafaelfranca:
I disagree with that. We just did the opposite change on rails/rails and it is working very well. Having the lockfile in the repository make easier to contributors run bisects to find commits that introduced bugs. Also it make sure that everyone that clones your project get the tests green.

see https://github.com/rails/rails/pull/18992

PR #38, opened 2015-06-25
Fix example for binding to element attributes

alexcoco:
@qq99 

The example shows binding attributes as events but they are not events and don't need `event-`.


qq99:
:+1:

PR #39, opened 2015-06-29
document Twine.shouldDiscardEvent

kevinhughes27:
This feature was not documented and I had to look in shopify core to figure out what it did. Figured a note in the readme was a good idea.

for review
@pushrax @qq99

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,18 @@ Example:
 <div context='bar' define='{bar: new Foo}'></div>
 ```
 
+## Twine.shouldDiscardEvent
+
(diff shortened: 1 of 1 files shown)


qq99:
So this is sort of like a blacklist?  Can you specify that all click handlers will have this function run before them?  Can you amend the copy to say, "If the function you set returns `true`, then the event processing chain will be halted" or something to that affect?

kevinhughes27:
yeah thats how it works. I'll make your change

kevinhughes27:
@pushrax you good with the wording here?

PR #40, opened 2015-06-29
document optimization

kevinhughes27:
I was just poking around learning about how Twine works and was implemented. Ending up focusing on this part a bit. I think a comment is worth it here to distinguish this as perf based choice rather than a logic one.

@pushrax @qq99


qq99:
nitpick: change your commit message to reflect that it's adding a comment, rather than optimizing something :P

kevinhughes27:
done!

pushrax:
Thanks!

PR #44, opened 2015-10-06
Support data attributes

### LINKED ISSUE #43: Allow `data-` attrs for valid HTML5
No page that uses Twine is valid HTML because of the custom attributes. I'm not suggesting killing these unprefixed versions, but it would be great if we could alias all of Twine’s attributes (e.g. `bind`, `define`) to data attributes (e.g. `data-bind`, `data-define`). That way any users that care about having valid markup can use the `data-` ones, and those that don’t can continue using the unprefixed versions.

nwtn:
Fixes #43.


lib/assets/javascripts/twine.js.coffee
```diff
@@ -326,5 +336,8 @@ setupEventBinding = (eventName) ->
     return teardown: ->
       $(node).off eventName, onEventHandler
 
+  Twine.bindingTypes["data-bind-event-#{eventName}"] = Twine.bindingTypes["bind-event-#{eventName}"]
```
DrewMartin: Probably cleaner to have a line at the end that does something like
```coffeescript
for key in Object.keys(Twine.bindingTypes)
  Twine.bindingTypes["data-#{key}"] = Twine.bindingTypes[key]
```

And then have a test that ensures that for every `bind-whatever

qq99:
cc @Shopify/tnt

dist/twine.js
```diff
@@ -307,8 +307,8 @@ wrapFunctionString = function(code, args, node) {
   } else {
     try {
       return new Function(args, "with($context) { return " + code + " }");
-    } catch (_error) {
-      e = _error;
+    } catch (error) {
+      e = error;
```
qq99: do we ever use `e` or `error`?

lib/assets/javascripts/twine.js.coffee
qq99: :+1:

dist/twine.js
celsodantas: I wonder why this is being renamed.

DrewMartin:
Oh you know what? This will lead to subtle bugs. We kind of depend on the order of the keys in `Twine.bindingTypes` so that when we go over the bindings on a node (https://github.com/Shopify/twine/blob/data-attributes/lib/assets/javascripts/twine.js.coffee#L48-L51) we do them in a specific order. If someone were to use a mix of data and non-data attributes, say `<input type="text" bind-event-input="onInput()" data-bind="val">` it would not work like you expect because the input event triggering `onInput()` would be fired before the input event that changes `val` so inside of the `onInput` method it would have the old value when you look at `@val`. #31 had the same issue (fixed in #33)

dist/twine.js
pushrax: CoffeeScript doing CoffeeScript things.

DrewMartin:
Alternative approach: change line 48 to
`for type, binding of Twine.bindingTypes when definition = (node.getAttribute(type) || node.getAttribute("data-#{type}"))`

pushrax:
Approach @DrewMartin gave is probably the best one, despite the associated performance cost. It might be good to have this feature enabled by configuration.

qq99:
Good catch @DrewMartin

nwtn:
Great feedback, thanks. I've updated with the suggested change from @DrewMartin.

nwtn:
ping @DrewMartin @pushrax @qq99

DrewMartin:
:+1:

qq99:
LGTM

nwtn:
cool cool cool. is there anything that needs to be done besides merging and following the [release steps](https://github.com/Shopify/twine#releasing)? e.g. shipit?

qq99:
iirc, that's all

I thought we were listed on bower at one point (https://github.com/Shopify/twine/blob/master/bower.json) but I am not seeing us on the registry

PR #45, opened 2015-11-05
correct js demo in readme.md

tessalt:
linting yells at me every time I copy and paste the boilerplate code :anguished:

Also made it more explicit that `context` is attached to `window` and someone didn't just forget to write `var`. 

@Shopify/tnt

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,9 +31,9 @@ Twine can be initialized simply with the following:
 
 ```html
 <script type="text/javascript">
-  context = {}
+  var context = {};
   $(function() {
-    Twine.reset(context).bind().refresh()
+    Twine.reset(context).bind().refresh();
   });
 </script>
 ```


qq99:
I think it was intended to be `var context` s.t. you can never access it via `window.context` again and muddle with it (to force you to go through `Twine.context(node)`).  The perils of manually transpiling from coffee to js :)

Then again, I don't think it really matters for a readme whether it's on `window` or not.  I think it may encourage more encapsulation if you don't have global access to the root context?  I have no objections to merging this

tessalt:
@qq99 updated to `var context` because I agree with your reasoning there.

qq99:
:+1:

tessalt:
related PR for example page: https://github.com/Shopify/twine/pull/46

DrewMartin:
:+1:

PR #47, opened 2015-11-12
Remove jQuery dependency

tmlayton:
I think this is all that is needed to remove Twine's dependency on jQuery. Here is a [codepen of the demo page without jQuery](http://codepen.io/timlayton/pen/EVyBKK) using the changes here

**Browser support for `Element.classList`:**

screen shot 2015-11-14 at 7 27 43 pm
screen shot 2015-11-14 at 7 27 54 pm

@Shopify/tnt @Shopify/fed 

### Todo
- [ ] Add tests using native events for where custom events are used
- [ ] Polyfill `classList`


tessalt:
nice. you'll also want to remove all the `$` from the readme probably

tmlayton:
yep will do if this work out

bspellio:
Hey guys, been watching this project for a while now. I love the idea of dropping this dependency. But, keep in mind that the second argument for `classList.toggle()` [isn't supported at all by IE](https://developer.mozilla.org/en/docs/Web/API/Element/classList) Perhaps if you wrote that logic yourself and just used `classList.toggle()` with a single argument, which is supported by IE 10+.

qq99:
This looks good to me, aside from the `classList` compatibility thing 

Off-hand I can't think of anything that'd go wrong

DrewMartin:
Looks like legit test failures

tmlayton:
@bspellio good catch on the browser support for `toggle`'s second argument, thanks. Pushed an update

test/twine_test.coffee
```diff
@@ -277,10 +277,10 @@ suite "Twine", ->
       node = setupView(testView, context = fn: @spy())
       data = {test: 'bla123'}
 
-      $(node).trigger 'submit', data
+      triggerEvent node, 'submit', data
 
       assert.isTrue context.fn.calledOnce
-      assert.isTrue context.fn.calledWith(data)
+      assert.isTrue context.fn.calledWith(data), 'fn called with data'
```
tmlayton: Can't get this assertion to pass, anyone able to help?

lib/assets/javascripts/twine.js.coffee
```diff
@@ -274,7 +280,11 @@ Twine.bindingTypes =
     return refresh: ->
       newValue = fn.call(node, context, rootContext)
       for key, value of newValue when lastValue[key] != value
-        $(node).attr(key, value || null)
+        if value
+          value = value.call() if typeof value == 'function'
```
tmlayton: Does this look okay? The jQuery `attr()` method allows for the second argument to be a function. This gets the test to pass but not sure if we should allow for the function to have arguments

tmlayton:
What is Twine's current browser support? If using `Element.classList` drops support for a version or more of IE should we bump the major version to 1?

lib/assets/javascripts/twine.js.coffee
qq99: I'm uncertain that there could be an occurence of `value` being a function that needs arguments.  I guess if you did `bind-placeholder="someFunction(foo, bar, baz)"` and `someFunction()` itself returned yet another function, then `value` could be a function for `$.attr` to use.  Am I correct in understanding this is the scenario at hand?

In the jQuery impl, I don't think it could ever call that fn with any arguments, since the destructuring into `key, value` there would yield 1 key and 1 single object for `value`.

I think - tentatively - you're OK with not allowing arguments since there should be no way to pass them through in the jQ implementation.

test/twine_test.coffee
qq99: Maybe try a debugger to look at `context.fn.args` (if I remember my sinon/chai properly) to see what it was indeed called with?  I'm guessing no args

qq99: https://github.com/Shopify/twine/pull/47/files#diff-54cce0c8766de7686ed6f38e8fe0c71dR528 is another good place to debugger

test/twine_test.coffee
```diff
@@ -6,9 +6,9 @@ suite "Twine", ->
     Twine.reset(context, rootNode).bind().refreshImmediately()
     rootNode.children[0]
 
-  triggerEvent = (node, eventName) ->
-    event = document.createEvent("HTMLEvents")
-    event.initEvent eventName, false, true
+  triggerEvent = (node, eventName, data = {}) ->
+    event = document.createEvent('CustomEvent')
+    event.initCustomEvent eventName, false, true, data
```
qq99: https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/initCustomEvent cites deprecation warning

May be related to your data failing to pass thru in the test below?

test/twine_test.coffee
```diff
@@ -277,10 +277,10 @@ suite "Twine", ->
       node = setupView(testView, context = fn: @spy())
       data = {test: 'bla123'}
 
-      $(node).trigger 'submit', data
```
qq99: While the lib shouldn't use jQ, I think it's fair game to use it in the tests if that makes the tests pass.  Does your troublesome test pass if you use jQ as the triggerer?

qq99:
>What is Twine's current browser support?

We're now using it on our brochure pages, which support considerably lower versions of IE.

test/twine_test.coffee
tmlayton: The test fails using jQ trigger too. I rewrote `triggerEvent` to try and get the test to pass

qq99: I was thinking last night that it may be a good idea to have tests that trigger the event from a variety of methodologies, to ensure it's working

e.g.,

`domEl.click()` on the DOM element itself
`triggerEvent`
`$(domEl).trigger('click')` or `$(domEl).click()` 

If it fails with jQ trigger, it seems to imply then that it's not the triggering but the handler for the event

qq99: Apparently, jQuery only really triggers listeners (when using `$.trigger`) on events that jQuery has itself bound...  which really sucks to anyone trying to trigger `addEventListener` events via `$.trigger`

This test is a bit weird in that seeing `bind-event-submit="fn(data)"`implies to me (at first glance) that `data` should be existing in the current Twine context, but the previous jQuery implementation was passing `data` through with the event.  So it seems more like a test of `$.trigger(eventname, data_pojo)`.

The rest of the event handler implementation inside Twine - `wrapFunctionString(definition, '$context,$root,event,data', node).call(node, context, rootContext, event, data)` - makes the presence of the 2nd data param into a keyword of sorts that any event handler can access, I guess officially authorizing firing Twine events thru `$(node).trigger("eventname", some_pojo)`

I did a search in our erb files for `(data` and there's a few binds with this signature... and more than a few calls from coffee files where we do `$(node).trigger` with a 2nd data argument.  Perhaps an argument can be made to remove the allowance of this behaviour, but it's unintuitive to any consumer and I would suggest a `1.0.0` release.  

Maybe this is an argument to add a `Twine.trigger`?  `Twine.trigger(node, "event_name", data_pojo)` or something which dispatches the `CustomEvent` as below in the tests

tmlayton:
It looks like `$.trigger` will fire events bound with `addEventListener`, although there seems to be an issue with the `'submit'` event specifically where `$(node).trigger 'submit'` and `node.submit()` do not fire the listener, yet `node.dispatchEvent customSubmitEventObj` will fire the listener... 

I feel like I'm missing something obvious: http://codepen.io/timlayton/pen/d2baa16dc28fc19702daca3b490baeea/

And how were these tests passing before?

qq99:
I assume `$.trigger` with `submit` works on `$.on` with `submit`?

tmlayton:
Yes, you are correct. `$.trigger 'submit'` will fire the listener if bound with `$.on 'submit'`.

qq99:
Some musing:

What if we made Twine use jQuery if Twine detects that jQuery is defined, otherwise it falls back to using the pure DOM methods added here?  It seems like it'd be the easiest way to bump Twine without impacting any consumers doing event triggering.  This implementation would make it so that we don't need to have `1.0.0` and a `1.0.0-jquery` releases running parallel to one another.  Further, we could have a `Twine.config` function to explicitly call out whether or not you want Twine to be using jQuery for eventing.

I think it could still be nice (at some point in the future? I don't know that it has to be in this PR) to have a `Twine.trigger(node, eventname, data)` function as our recommended methodology of triggering twine-bound events.  I guess there's still some concern with triggering as http://codepen.io/timlayton/pen/d2baa16dc28fc19702daca3b490baeea/ demonstrates

qq99:
@tmlayton odd: if you try to use `formEl.onsubmit = function (event) {...}` then jQuery incarnations of triggering submit seem to actually trigger it.

tmlayton:
@qq99 (and anyone else who wants to weigh in) perhaps 3 options, maybe others:

**Option 1**: if the `submit` event is the only issue then we could conditionally bind using `onsubmit`. Do you know if any other events that `$.trigger` won't fire is bound using addEventListener? Bump a minor version

**Option 2**: detect jQuery and conditionally use event binders. Bump a minor version

**Option 3**: Leave this as-is and add `Twine.trigger` in this PR. Shopify will just need to update any uses of `$.trigger`. Release as a major version

And I think in any of these options we will polyfill `classList`

qq99:
I am biased to option 2

tessalt:
I also am partial to option 2

tmlayton:
Going to close this for now, between jQuery detection and the `classList` polyfill it feels meh. I'd rather wait until we can cleanly drop <= IE 9 support.

PR #51, opened 2015-12-10
fixed data attributes

nwtn:
@tylerball noticed that `data-context` wasn’t working. This PR allows `data-` prefixes on `context` and `allow-default`, which were missed last time around, and adds more robust tests for `data-` prefixed attributes.

I noticed there are no tests for `allow-default`; is anybody able to write some that test the intended behaviour? (I could probably come up with one if nobody else is game.)

@Shopify/tnt

### DIFF
--- a/lib/assets/javascripts/twine.js.coffee
+++ b/lib/assets/javascripts/twine.js.coffee
@@ -17,6 +17,9 @@ rootNode = null
 
 currentBindingCallbacks = null
 
+Twine.getAttribute = (node, attr) ->
+  node.getAttribute("data-#{attr}") || node.getAttribute(attr)
+
 # Cleans up all existing bindings and sets the root node and context.
 Twine.reset = (newContext, node = document.documentElement) ->
   for key of elements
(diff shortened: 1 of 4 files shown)


pushrax:
It's starting to look like this logic should be abstracted into another function. Otherwise LGTM.

nwtn:
yup, fair point. abstracted this out to a `Twine.getAttribute` method. :eyes: @Shopify/tnt

qq99:
quickly :eyes:, :+1:

tmlayton:
Tests have been added for `data-allow-default` and `allow-default`

lib/assets/javascripts/twine.js.coffee
```diff
@@ -17,6 +17,9 @@ rootNode = null
 
 currentBindingCallbacks = null
 
+Twine.getAttribute = (node, attr) ->
```
tmlayton: Nice

tmlayton:
:+1:

nwtn:
cool. tim’s tests lgtm and everything is green. merging.

PR #54, opened 2015-12-21
0.1.4

nwtn:
@celsodantas @tmlayton  @Shopify/tnt

i had already updated the version in `lib/twine-rails/version.rb` but forgot to do so in `package.json` and `bower.json`. this does that, along with a `bundle install` and `make .all && make .uglify`.


tmlayton:
Seems reasonable, I haven't done a release of Twine though

celsodantas:
For reference https://github.com/Shopify/twine/pull/25/files

celsodantas:
:+1:

qq99:
:+1:

Issue #53, opened 2015-12-21
bind-class doesn't behave like other bind-title, placeholder, href, etc


drewhamlett:
If do a bind-title the expected value is shown, but if i do bind-class it does not work as expected.
```
            <%= link_to('#', class: 'btn btn-purple btn-icon',
                             'bind-event-click' => "addNoteToggle()",
                             'bind-title' => 'note.visible',
                             'bind-class' => '{ active: note.visible }') do %>
```
After I submit a remote form a SJR comes back and does a Turbolinks replace on a section.  Everything is then reset via

```
$(document).on('page:partial-load', function() {
  console.log('Partial Load')
  Twine.reset(window.context).bind().refresh()
})```

If I print out the current context, note.visible comes back false, as does bind-title.  bind-class is not changed.
```
  'bind-class': (node, context, definition) ->
    fn = wrapFunctionString(definition, '$context,$root', node)
    lastValue = {}
    return refresh: ->
      newValue = fn.call(node, context, rootContext)
      for key, value of newValue when !lastValue[key] != !value
        $(node).toggleClass(key, !!value)
      lastValue = newValue
```

If I take out ```when !lastValue[key] != !value``` from bind class everything works as expected.

pushrax:
So this is a bug right? The default `lastValue[key]` is `undefined` so the first time it runs `!lastValue[key] == !value` since `!undefined == !false`.

I'm not sure totally what you mean about your test though – since the node doesn't have the `active` class manually defined it doesn't seem like it would be a problem (`toggleClass` isn't called, and since the node doesn't have the class that's fine).

PR #55, opened 2016-01-09
performance/reliability updates to bind-class function

SkylerRogers:
fixes issue #53


SkylerRogers:
fixes issue #53

pushrax:
Is this really the correct fix? We don't want to call this potentially expensive function every time a refresh is triggered. Do you know why the condition was incorrect?

pushrax:
OK, so I think the issue is because `!undefined === true`. It seems that I didn't expect there to be cases where a class needs to be initially removed from a node in the first refresh. I think the correct way to deal with this is to load the current classes from the node when it's being bound, and store them into the `lastValue` object.

SkylerRogers:
I hear what you're saying. My thinking was that class binds are typically a one way data flow and that the performance loss from always calling toggleClass would somewhat "net itself out" from the removal of the constant equality checks. I agree, it's not perfect nor fully calculated but, at least on my end, appears to be more reliable approach in terms of reducing race conditions and the need to force manipulate change events to get Twine to update the views.

I'm still learning the Twine architecture though so perhaps your solution is better than mine. Happy to try your version.

Thanks for helping me look into this,
-Skyler

pushrax:
DOM operations are very expensive, even if they do nothing, as they need to call into C and do various other things, whereas this equality check in JS is basically free (a couple jitted instructions). This is the same approach taken frameworks like AngularJS, and part of the reason why the fancy virtual DOM and diffing algorithm in React exists.

On Monday, I can take a look at implementing the solution I described, or you can feel free to implement it yourself on this branch :smile:

SkylerRogers:
Sorry didn't realize you were concerned about DOM updates specifically. If thats the case my current solution may be no good. Additionally, shouldn't this bind-class function also make a single jQuery toggleClass call with all of the updated classes as opposed to calling toggleClass per class/key!? Or at least one for removals and one for additions?

Thanks,
-Skyler

SkylerRogers:
What about something like this:
    
    value = !!newValue[key];
    if($(node).hasClass(key) !== value){
      $(node).toggleClass(key, value);
    }

It gets rid of the fragile last value tracker but isn't updating the DOM more than necessary?

pushrax:
Great idea to batch all the removals and additions! I ran some benchmarks, which showed there is a significant improvement when there are multiple classes to be added/removed, and no performance reduction when there's only one class.

http://jsperf.com/multiple-toggleclass-vs-single-batch
http://jsperf.com/single-toggleclass-vs-single-batch

I'd like to avoid querying the DOM on each refresh, as there isn't any reason why the cached state should be incorrect. I also realized that loading all the classes of every node at bind-time (my first suggestion) could increase memory usage by a fair amount - so I think the right way to approach this is to use `hasClass` the first time `refresh` runs, and then use the cache every time after that.

SkylerRogers:
It might be able to be DRYed up a little, but maybe something like this?

     'bind-class': function(node, context, definition) {
        var fn, lastValue;
        fn = wrapFunctionString(definition, '$context,$root', node);
        lastValue = {};
        return {
          refresh: function() {
            var key, newValue, value, cached, additions, removals;
            newValue = fn.call(node, context, rootContext);
            additions = [];
            removals = [];
            for (key in newValue) {
              value = newValue[key] = !!newValue[key];
              cached = lastValue[key];
              if (cached === undefined) {
                if ($(node).hasClass(key) !== value) {
                  value ? additions.push(key) : removals.push(key);
                }
              } else {
                if (cached !== value) {
                  value ? additions.push(key) : removals.push(key);
                }
              }
            }
            if (removals.length) {
              $(node).removeClass(removals.join(" "));
            }
            if (additions.length) {
              $(node).addClass(additions.join(" "));
            }
            return lastValue = newValue;
          }
        };
      }

SkylerRogers:
So how exactly does the cache work? It looks like there is an intent here to

SkylerRogers:
Sorry didn't mean to close

SkylerRogers:
bump

lib/assets/javascripts/twine.js.coffee
```diff
@@ -268,10 +268,19 @@ Twine.bindingTypes =
   'bind-class': (node, context, definition) ->
     fn = wrapFunctionString(definition, '$context,$root', node)
     lastValue = {}
+    additions = []
+    removals = []
```
pushrax: Since `$(node)` is used several times, may as well store it at the start in a variable (e.g. `$node`).

lib/assets/javascripts/twine.js.coffee
```diff
@@ -268,10 +268,19 @@ Twine.bindingTypes =
...
     return refresh: ->
       newValue = fn.call(node, context, rootContext)
-      for key, value of newValue when !lastValue[key] != !value
-        $(node).toggleClass(key, !!value)
+      for key, value of newValue
+        value = newValue[key] = !!newValue[key]
```
pushrax: Don't need to update `newValue[key]`

lib/assets/javascripts/twine.js.coffee
```diff
@@ -268,10 +268,19 @@ Twine.bindingTypes =
...
+        cached = lastValue[key]
+        if cached == undefined and $(node).hasClass(key) != value or cached != value
```
pushrax: I would consider replacing the `cached == undefined and $(node).hasClass(key) != value` part with `cached ?= $node.hasClass(key)` on the previous line.

lib/assets/javascripts/twine.js.coffee
```diff
@@ -268,10 +268,19 @@ Twine.bindingTypes =
...
+          if value then additions.push(key) else removals.push(key)
+      if removals.length
+        $(node).removeClass removals.join(' ')
+      if additions.length
+        $(node).addClass additions.join(' ')
```
pushrax: Either use the postfix notation (`if` after the statement) or add some newlines. Convention in this codebase is also to use brackets for function calls unless the argument spans multiple lines.

pushrax:
This is exactly what I was thinking, awesome. Just some stylistic comments and then this is ready to be merged.

lib/assets/javascripts/twine.js.coffee
SkylerRogers: even though the newValue eventually gets cached?

lib/assets/javascripts/twine.js.coffee
SkylerRogers: good call

lib/assets/javascripts/twine.js.coffee
SkylerRogers: Sorry about that (JS to coffee isn't the cleanest)

lib/assets/javascripts/twine.js.coffee
SkylerRogers: Sorry not entirely following you here on syntax preference (BTW this was the part to handle the initial cache value being undefined)

SkylerRogers:
I pushed some updates. I made some comments after yours as well (please click 'show outdated diff' to view)

Many thanks,
-Skyler

TheMallen:
Is this still good if we resolve the conflicts?

SkylerRogers:
@TheMallen - I rebased with the latest master and re-applied my original fix. The tests are passing but its been a while since I've been in here so please give it an extra look just in case.

Cheers,
-Skyler

lib/assets/javascripts/twine.coffee
```diff
@@ -364,10 +364,23 @@
...
+        removals = []
+        for key, value of newValue
+          value = newValue[key] = !!newValue[key]
+          cached = lastValue[key]
+          if cached == undefined && $node.hasClass(key) != value || cached != value
+            if value
+              additions.push(key)
+            else
+              removals.push(key)
+        if removals.length
```
TheMallen: If you change this to `$node.removeClass(removals.join(' ')) if removals.length` I think this looks good.

lib/assets/javascripts/twine.coffee
```diff
@@ -364,10 +364,23 @@
...
+          $node.removeClass(removals.join(' '))
+        if additions.length
```
TheMallen: ` $node.addClass(additions.join(' ')) if additions.length` here as well

TheMallen:
@pushrax do you think this is good to go if the ifs get postfixed?

pushrax:
Thanks, and sorry this took so long!

PR #58, opened 2016-03-15
add bind-prop

richardmonette:
##### Background

Turns out that properties and attributes differ in that an attribute is the initial value whereas the property is the dynamic value.

| $ accessor | return value |
|---|------|
| elem.checked  | (Boolean) Will change with checkbox state |
| $( elem ).prop( "checked" )  |  (Boolean) Will change with checkbox state    |
| elem.getAttribute( "checked" )	  |  (String) Initial state of the checkbox; does not change    |

for more info, see 'Attributes vs. Properties' here: http://api.jquery.com/prop/

##### Why add this?

This would generally be a nice thing to have, but specifically it would allow for binding properties of https://styleguide.myshopify.com/components/form UI components such as `indeterminate` on checkboxes. 

@qq99 @pushrax @maartenvg @katdrobnjakovic

### DIFF
--- a/lib/assets/javascripts/twine.js.coffee
+++ b/lib/assets/javascripts/twine.js.coffee
@@ -283,6 +283,15 @@ Twine.bindingTypes =
         $(node).attr(key, value || null)
       lastValue = newValue
 
+  'bind-prop': (node, context, definition) ->
+    fn = wrapFunctionString(definition, '$context,$root', node)
+    lastValue = {}
+    return refresh: ->
+      newValue = fn.call(node, context, rootContext)
+      for key, value of newValue when lastValue[key] != value
+        node[key] = value
+      lastValue = newValue
+
   define: (node, context, definition) ->
     fn = wrapFunctionString(definition, '$context,$root', node)
     object = fn.call(node, context, rootContext)
--- a/test/twine_test.coffee
+++ b/test/twine_test.coffee
@@ -196,6 +196,23 @@ suite "Twine", ->
       node = setupView(testView, key: true)
       assert.equal node.className, "cls"
 
+  suite "data-bind-prop attribute", ->
+    test "should apply the given prop when truthy", ->
+      testView = "<div data-bind-prop=\"{a: true_key, b: false, c: 0, d: null, e: undefined, f: string_key}\"></div>"
+      node = setupView(testView, true_key: true, string_key: "yolo")
+      assert.equal node.a, true
+      assert.isTrue node.hasOwnProperty('a')
+      assert.equal node.b, false
+      assert.isTrue node.hasOwnProperty('b')
+      assert.equal node.c, 0
+      assert.isTrue node.hasOwnProperty('c')
+      assert.equal node.d, null
+      assert.isTrue node.hasOwnProperty('d')
+      assert.equal node.e, null
+      assert.isFalse node.hasOwnProperty('e')
+      assert.equal node.f, 'yolo'
+      assert.isTrue node.hasOwnProperty('f')
+
   suite "data-bind-attribute attribute", ->
     test "should apply the given attribute when truthy", ->
       testView = "<div data-bind-attribute=\"{a: key, b: false, c: 0, d: null, e: undefined, f: function() { return true }, g: '0', h: 'false'}\"></div>"
--- a/dist/twine.js
+++ b/dist/twine.js
@@ -301,7 +301,7 @@ stringifyNodeAttributes = function(node) {
 };
 
 wrapFunctionString = function(code, args, node) {
-  var e, error, keypath;
+  var e, keypath;
   if (isKeypath(code) && (keypath = keypathForKey(code))) {
     if (keypath[0] === '$root') {
       return function($context, $root) {
@@ -315,8 +315,8 @@ wrapFunctionString = function(code, args, node) {
   } else {
     try {
       return new Function(args, "with($context) { return " + code + " }");
-    } catch (error) {
-      e = error;
+    } catch (_error) {
+      e = _error;
       throw "Twine error: Unable to create function on " + node.nodeName + " node with attributes " + (stringifyNodeAttributes(node));
     }
   }
@@ -442,6 +442,24 @@ Twine.bindingTypes = {
           value = newValue[key];
           if (lastValue[key] !== value) {
             $(node).attr(key, value || null);
+          }
+        }
+        return lastValue = newValue;
+      }
+    };
+  },
+  'bind-prop': function(node, context, definition) {
+    var fn, lastValue;
+    fn = wrapFunctionString(definition, '$context,$root', node);
+    lastValue = {};
+    return {
+      refresh: function() {
+        var key, newValue, value;
+        newValue = fn.call(node, context, rootContext);
+        for (key in newValue) {
+          value = newValue[key];
+          if (lastValue[key] !== value) {
+            node[key] = value;
           }
         }
         return lastValue = newValue;


qq99:
cc @Shopify/tnt

pushrax:
I think rather than building this in, we should consider removing jQuery. Before discussing the actual PR, what do people think of that?

test/twine_test.coffee
```diff
@@ -196,6 +196,21 @@ suite "Twine", ->
       node = setupView(testView, key: true)
       assert.equal node.className, "cls"
 
+  suite "data-bind-prop attribute", ->
+    test "should apply the given prop when truthy", ->
+      testView = "<div data-bind-prop=\"{a: key, b: false, c: 0, d: null, e: undefined}\"></div>"
+      node = setupView(testView, key: true)
+      assert.equal node.a, true
+      assert.isTrue node.hasOwnProperty('a')
+      assert.equal node.b, null
```
qq99: shouldn't this be `false`?

nwtn:
> I think rather than building this in, we should consider removing jQuery. Before discussing the actual PR, what do people think of that?

@tmlayton tried that and ran into a heap of complications. I think it’s a laudable goal, though.

test/twine_test.coffee
```diff
@@ -196,6 +196,21 @@ suite "Twine", ->
       node = setupView(testView, key: true)
       assert.equal node.className, "cls"
 
+  suite "data-bind-prop attribute", ->
+    test "should apply the given prop when truthy", ->
+      testView = "<div data-bind-prop=\"{a: key, b: false, c: 0, d: null, e: undefined}\"></div>"
+      node = setupView(testView, key: true)
+      assert.equal node.a, true
```
qq99: and this, the value of `key`?

test/twine_test.coffee
```diff
@@ -196,6 +196,21 @@ suite "Twine", ->
...
+      assert.isTrue node.hasOwnProperty('a')
+      assert.equal node.b, null
+      assert.isTrue node.hasOwnProperty('b')
+      assert.equal node.c, null
```
qq99: not `0`?

richardmonette:
> I think rather than building this in, we should consider removing jQuery. Before discussing the actual PR, what do people think of that?

@pushrax I support it in principle, but honestly with the time sensitivity of what I need this for, I really don't want to see this get side tracked.

nwtn:
@richardmonette would you be able to spin up a codepen that shows this in use?

test/twine_test.coffee
richardmonette: value of `key` is true... (I think?)

pushrax: well, the value is `true`

test/twine_test.coffee
richardmonette: Setting a property to any falsy value removes it, which makes it null.

qq99:
I support pushing this in now, and trying again at jQuery at a later date

test/twine_test.coffee
qq99: hah :)
maybe add one w/ a string?

pushrax:
The trouble is that adding this now will make the jQuery :fire: more difficult later. If you need it urgently then we can deal with that though.

test/twine_test.coffee
```diff
@@ -196,6 +196,21 @@ suite "Twine", ->
...
+      assert.isTrue node.hasOwnProperty('c')
+      assert.equal node.d, null
+      assert.isTrue node.hasOwnProperty('d')
+      assert.equal node.e, null
+      assert.isFalse node.hasOwnProperty('e')
```
qq99: setting to `undefined` similarly removes / never adds the property?

lib/assets/javascripts/twine.js.coffee
```diff
@@ -283,6 +283,15 @@ Twine.bindingTypes =
         $(node).attr(key, value || null)
       lastValue = newValue
 
+  'bind-prop': (node, context, definition) ->
+    fn = wrapFunctionString(definition, '$context,$root', node)
+    lastValue = {}
+    return refresh: ->
+      newValue = fn.call(node, context, rootContext)
+      for key, value of newValue when lastValue[key] != value
+        $(node).prop(key, value || null)
```
DrewMartin: Regarding `value || null`: not sure if this if this is what the behaviour should be. For `bind-attribute` it makes sense because setting, say, the `disabled` attribute to `false` will make it look like `disabled="false"` and still be disabled, when you'd want it to enabled. For checked, at least, false will uncheck it the same as null would.

DrewMartin: Hmm though trying to set a value to `undefined` will actually be equivalent to calling `$(el).prop('checked')` which just returns the property

DrewMartin: So maybe this should be `value ? null` so if it's evaluated to undefined it will instead set it to null, but other values that may be relevant (0, false) will be preserved

qq99: what about setting it via `node[key] = value` style? not related to @DrewMartin's comment, but moreso from the jQuery perspective? just throwing that out there

richardmonette:
@nwtn http://codepen.io/anon/pen/dMOqON

lib/assets/javascripts/twine.js.coffee
richardmonette: `node[key] = value || null` seems to work just as well, happy to change to that

nwtn:
Cool, thanks @richardmonette. Once the questions about the tests are resolved, I’m :+1:

lib/assets/javascripts/twine.js.coffee
DrewMartin: `node[key] = value` though

richardmonette:
OK, so not using `$.prop` and tests updated accordingly.

DrewMartin:
:+1:

pushrax:
So, just to confirm, you can't use `bind-indeterminate` in this case? In fact where do we need `bind-attribute` that `bind-<x>` doesn't work? I assume due to a module returning a large set of attributes to bind?

Can we change `bind-attribute` to perform the functionality in this PR instead?

DrewMartin:
>  In fact, where do we need bind-attribute that bind-<x> doesn't work?

`bind-disabled` isn't a thing

pushrax:
> `bind-disabled` isn't a thing

I assume you mean `bind-attribute`, and if you do then we should just add it [here](https://github.com/Shopify/twine/blob/5240a02d9bb8bd3a3e5e5fbfb68a77e64ab3afc8/lib/assets/javascripts/twine.js.coffee#L298).

DrewMartin:
Oh but apparently doing `el.disabled = true` actually sets the property. `tg-remote-noserialize` is one where `el['tg-remote-noserialize'] = true` wouldn't work

richardmonette:
bind-attribute has been used in the past for adding non-standard attribute such as `tg-remote-noserialize` to DOM nodes (which prevents serialization of the children) nodes.

pushrax:
Just to make sure I'm on the same page:

- `node[x]` is a property
- `node.attributes[x]` is an attribute

I don't believe there is another way to set `disabled` than through `node.disabled`.

I see what you mean about custom attributes names though. I'd rather have `bind-attributes` than dynamically created binding attributes.

qq99:
I thought `disabled` could be set via attribute?

maartenvg:
> Going back to this PR though, can we just add indeterminate to the list of properties?

Agreed. It's weird having both a generic and a list of specific prop-setters. So unless we deprecate most or all currently defined properties, there's not much we gain by adding a generic solution.

richardmonette:
Closing in favour of https://github.com/Shopify/twine/pull/61

PR #61, opened 2016-03-16
add indeterminate to augment checked

richardmonette:
Investigated adding `bind-prop` for this https://github.com/Shopify/twine/pull/58, but I think since this is roughly equivalent to `checked` we just add this as another option.

@qq99 @pushrax @maartenvg @katdrobnjakovic @Shopify/tnt

### DIFF
--- a/test/twine_test.coffee
+++ b/test/twine_test.coffee
@@ -209,6 +209,24 @@ suite "Twine", ->
       assert.equal node.getAttribute('g'), '0'
       assert.equal node.getAttribute('h'), 'false'
 
+  suite "data-bind-indeterminate attribute", ->
+    test "should set the indeterminate attribute", ->
+      testView = "<input type=\"checkbox\" data-bind-indeterminate=\"key\">"
+      node = setupView(testView, context = key: true)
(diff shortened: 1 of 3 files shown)


pushrax:
LGTM

test/twine_test.coffee
```diff
@@ -209,6 +209,16 @@ suite "Twine", ->
       assert.equal node.getAttribute('g'), '0'
       assert.equal node.getAttribute('h'), 'false'
 
+  suite "data-bind-indeterminate attribute", ->
+    test "should set the indeterminate attribute", ->
+      testView = "<input type=\"checkbox\" data-bind-indeterminate=\"key\">"
+      node = setupView(testView, context = key: true)
```
DrewMartin: Since it's a boolean attribute it should work with values other than strictly true and false. Can you test with other truthy/falsy values, particularly null and undefined?

richardmonette: Added null and undefined.

DrewMartin:
:ship:

maartenvg:
:shipit:

qq99:
LGTM

Issue #60, opened 2016-03-16
Binds on textContent crash IE when there is a mutationObserver on the page


TheMallen:
At the moment, `bind` is crashing IE when used on a page that also has a mutation observer (such as from an element query).

[Here](http://connect.microsoft.com/IE/feedbackdetail/view/851144/assigning-textcontent-crashes-the-browser-tab) is the IE bug.

[Here](https://jsfiddle.net/x2r3jpuz/2/embedded/result/) is a fiddle where you can crash IE by clicking a button. Wooo!

Possible solution: bind to innerText? :(

pushrax:
`textContent` and `innerText` do different things, so I don't think it's a viable solution. I don't really have an alternate solution though.

TheMallen:
A workaround is to bind to `innerHTML` for tags like `p` and then do manual escaping, but that is way sketchier. :(

PR #62, opened 2016-03-16
Add binding for innerText

dfmcphee:
This adds the ability to bind a value to the `innerText` property of a node. The default bind uses `textContent` which can [crash the browser](http://connect.microsoft.com/IE/feedbackdetail/view/851144/assigning-textcontent-crashes-the-browser-tab) in IE when there are too many bindings on a page.

@Shopify/tnt

### DIFF
--- a/lib/assets/javascripts/twine.js.coffee
+++ b/lib/assets/javascripts/twine.js.coffee
@@ -312,6 +312,7 @@ for attribute in ['placeholder', 'checked', 'disabled', 'href', 'title', 'readOn
   setupAttributeBinding(attribute, attribute)
 
 setupAttributeBinding('innerHTML', 'unsafe-html')
+setupAttributeBinding('innerText', 'inner-text')
 
 preventDefaultForEvent = (event) ->
   (event.type == 'submit' || event.currentTarget.nodeName.toLowerCase() == 'a') &&
--- a/test/twine_test.coffee
+++ b/test/twine_test.coffee
@@ -250,11 +250,11 @@ suite "Twine", ->
       Twine.refreshImmediately()
       assert.isFalse node.readOnly
 
-  suite "data-bind-unsafe-html attribute", ->
-    test "should set the innerHTML of the node", ->
-      testView = "<div data-bind-unsafe-html=\"key\"></div>"
-      node = setupView(testView, key: "&amp;")
-      assert.equal node.innerHTML, "&amp;"
+  suite "data-bind-inner-text attribute", ->
+    test "should set the innerText of the node", ->
+      testView = "<div data-bind-inner-text=\"key\"></div>"
+      node = setupView(testView, key: "some text")
+      assert.equal node.innerHTML, "some text"
 
   suite "data-bind-src attribute", ->
     test "should set the src of the node", ->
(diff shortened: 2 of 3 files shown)


qq99:
Closes #60 ?

dfmcphee:
Ah yes, this addresses #60. I figured doing explicitly like this was safer than trying to modify the behaviour of `bind` everywhere.

test/twine_test.coffee
```diff
@@ -250,11 +250,11 @@ suite "Twine", ->
       Twine.refreshImmediately()
       assert.isFalse node.readOnly
 
-  suite "data-bind-unsafe-html attribute", ->
-    test "should set the innerHTML of the node", ->
-      testView = "<div data-bind-unsafe-html=\"key\"></div>"
-      node = setupView(testView, key: "&amp;")
-      assert.equal node.innerHTML, "&amp;"
```
qq99: accidental removal here?

dfmcphee: Whoops. Will add this back.

dfmcphee:
I added that test back. Any other comments @qq99?

qq99:
Code LGTM, I believe
I don't have any opinions on whether it's the correct/right thing to do, but I don't see a lot of harm in exposing this at all :)

I don't remember what we do for release; do we bump version & make the `.min.js` version inside the PR? Or did we historically do that after PR is merged?

lib/assets/javascripts/twine.js.coffee
```diff
@@ -312,6 +312,7 @@ for attribute in ['placeholder', 'checked', 'disabled', 'href', 'title', 'readOn
   setupAttributeBinding(attribute, attribute)
 
 setupAttributeBinding('innerHTML', 'unsafe-html')
+setupAttributeBinding('innerText', 'inner-text')
```
pushrax: If we do this, it should just be added to the list above, making the binding `bind-innerText`.

pushrax:
How would you use this? It's not really OK to use `innerText` for other browsers.

dfmcphee:
Hm, that's a good point @pushrax. I did test it in all browsers and it did work as expected, but checking [caniuse](http://caniuse.com/#feat=innertext) shows that it only works in the latest version of Firefox.

Do you think having a fallback for IE is a better solution? I worry that then the differences between how text is handled with `innerText` vs `textContent` will cause unexpected weirdness in IE.

dfmcphee:
An alternative approach could be to create a new `bind-safe-html` attribute that uses `innerHTML` under the hood but also escapes the content for you.

pushrax:
Interestingly jQuery doesn't have any workaround for [this](https://github.com/jquery/jquery/blob/b0b280cd61d1fa12b67bd723ac2c2fa91b92db01/src/manipulation.js#L325).

I'm still not sure how you'd use `bind-safe-html`, is there some single field this is crashing for? If so, can we use some other workaround to avoid the crash?

I definitely want to avoid `innerHTML` in the majority of cases, since it's significantly more expensive.

dfmcphee:
Yeah, I am able to recreate using jQuery as well. If you open this [codepen](http://codepen.io/dfmcphee/pen/pyRgRq?editors=1010) and press the button in IE 11, it will crash too.

Right now we have a single page that is crashing in IE. I was hoping to update that binding specifically to use this instead of `bind`.

dfmcphee:
We ended up working around this by using `bind-unsafe-html` where needed with a manual escape. I am going to close this as I don't think it is the right solution anyway.

PR #67, opened 2016-04-14
fix dependency versions by using railties which doesn't include rails 5 beta

richardmonette:
OK, so our gem is is https://rubygems.org/gems/twine-rails which takes any coffee-rails version > 0

coffee-rails will take https://rubygems.org/gems/coffee-rails/versions/4.1.1 any railties version railties < 5.1.x, >= 4.0.0

The issue is that railties lastest version which matches the 5 series takes actionpack and activesupport which is 5.0.0.3beta which pulls in https://rubygems.org/gems/railties/versions/5.0.0.beta3

what we want is to stay on the 4.X series for now, which makes the current latest version https://rubygems.org/gems/railties

This changes the gemspec to make sure for railties we stay less than the 5 series, so we keep picking up 4.X stuff until we want to go Rails 5.

This also ensures that we are matching the 4.X versions we are using in Shopify core currently

@celsodantas @qq99 cc @Shopify/tnt


celsodantas:
@rafaelfranca might be good person to take a look too

maartenvg:
If we need it to be between 4 and 5, we should do `~> 4.0`.

rafaelfranca:
Urgh! The dependencies are missing. This gem depends on railtie and activesupport but it is not explicitly marking them as dependecy. If coffee-rails change its dependencies this gem will be broken. I'll fix it.

PR #70, opened 2016-05-20
Improve allow-default behavior

asmockler:
Currently, `allow-default` checks for a value of `'1'` explicitly, which can be a bit confusing. This PR changes its functionality to work more like the `checked` attribute - as long as it is present and has any value besides `'false'`, it will allow default.

cc: @dan-menard @beefchimi @Shopify/tnt

### DIFF
--- a/dist/twine.js
+++ b/dist/twine.js
@@ -473,7 +473,8 @@
     }
     setupAttributeBinding('innerHTML', 'unsafe-html');
     preventDefaultForEvent = function(event) {
-      return (event.type === 'submit' || event.currentTarget.nodeName.toLowerCase() === 'a') && Twine.getAttribute(event.currentTarget, 'allow-default') !== '1';
+      var ref1;
+      return (event.type === 'submit' || event.currentTarget.nodeName.toLowerCase() === 'a') && ((ref1 = Twine.getAttribute(event.currentTarget, 'allow-default')) === 'false' || ref1 === false || ref1 === 0 || ref1 === (void 0) || ref1 === null);
     };
     setupEventBinding = function(eventName) {
       return Twine.bindingTypes["bind-event-" + eventName] = function(node, context, definition) {
--- a/lib/assets/javascripts/twine.coffee
+++ b/lib/assets/javascripts/twine.coffee
@@ -323,7 +323,7 @@
 
   preventDefaultForEvent = (event) ->
     (event.type == 'submit' || event.currentTarget.nodeName.toLowerCase() == 'a') &&
-    Twine.getAttribute(event.currentTarget, 'allow-default') != '1'
+    Twine.getAttribute(event.currentTarget, 'allow-default') in ['false', false, 0, undefined, null]
 
   setupEventBinding = (eventName) ->
     Twine.bindingTypes["bind-event-#{eventName}"] = (node, context, definition) ->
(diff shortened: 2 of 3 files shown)


dist/twine.js
```diff
@@ -499,7 +499,8 @@ for (j = 0, len = ref.length; j < len; j++) {
 setupAttributeBinding('innerHTML', 'unsafe-html');
 
 preventDefaultForEvent = function(event) {
-  return (event.type === 'submit' || event.currentTarget.nodeName.toLowerCase() === 'a') && Twine.getAttribute(event.currentTarget, 'allow-default') !== '1';
+  var ref1;
```
dan-menard: Might as well make this `ref1 = Twine.getAttribute(event.currentTarget, 'allow-default')`, right? Saves a bit of length on the next line, and the confusion of making an assignment inside a conditional.

asmockler: This got auto-compiled ¯\_(ツ)_/¯ but I don't think it is an unreasonable change to make.

lib/assets/javascripts/twine.js.coffee
```diff
@@ -315,7 +315,7 @@ setupAttributeBinding('innerHTML', 'unsafe-html')
 
 preventDefaultForEvent = (event) ->
   (event.type == 'submit' || event.currentTarget.nodeName.toLowerCase() == 'a') &&
-  Twine.getAttribute(event.currentTarget, 'allow-default') != '1'
+  Twine.getAttribute(event.currentTarget, 'allow-default') in ['false', false, undefined, null]
```
nwtn: 0?

asmockler:
@nwtn Good call on checking for 0 as well. Added it.

dist/twine.js
pushrax: Yeah this file shouldn't be hand-edited

pushrax:
Seems reasonable

asmockler:
@shopify/tnt Never got around to merging - everyone still good with this?

lib/assets/javascripts/twine.js.coffee
```diff
@@ -315,7 +315,7 @@ setupAttributeBinding('innerHTML', 'unsafe-html')
 
 preventDefaultForEvent = (event) ->
   (event.type == 'submit' || event.currentTarget.nodeName.toLowerCase() == 'a') &&
-  Twine.getAttribute(event.currentTarget, 'allow-default') != '1'
+  Twine.getAttribute(event.currentTarget, 'allow-default') in ['false', false, 0, undefined, null]
```
qq99: maybe we want to make a `Twine.hasAttribute` and use https://developer.mozilla.org/en-US/docs/Web/API/Element/hasAttribute ?

asmockler: We would still have to check for falsey values right? So that would end up looking like 

```
(event.type == 'submit' || event.currentTarget.nodeName.toLowerCase() == 'a') &&
Twine.hasAttribute(event.currentTarget, 'allow-default') &&
Twine.getAttribute(event.currentTarget, 'allow-default') in ['false', false, 0, undefined, null]
```

right?

qq99: Yeah I'm wondering if we have any spots where we're doing `allow-default="false"` falsey type things? Isn't this something people only add on after-the-fact to allow the default?

qq99: Not a big worry really, mostly yields a tiny perf benefit

I'm cool with this PR as written 👍

PR #71, opened 2016-05-20
Add dev.yml

asmockler:
Adds `dev.yml` to make it (slightly) easier to get up and running with Twine. 

@Shopify/tnt

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "coffeeify": "~0.7.0",
     "uglify-js": "~2.4.15",
     "browserify": "5.11.1",
-    "testem": "0.6.18",
+    "testem": "1.7.1",
     "coffee-script": "1.10.0"
   }
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -31,6 +31,6 @@
     "uglify-js": "~2.4.15",
     "browserify": "5.11.1",
     "pbind": "~0.0.1",
-    "testem": "0.6.18"
+    "testem": "1.7.1"
```
asmockler: Bumped the `testem` version because `0.6.18` would hang when running `testem ci`. Since the current instructions are just to `npm install -g testem`, I updated to the latest (since that's what `<-` that would do)

asmockler:
@Shopify/tnt Could I get 👀  on this?

asmockler:
Bumped to Ruby 2.3.1

PR #74, opened 2016-06-16
Adds Twine.register

tsov:
Twine currently relies on looking up modules/constructors within the global namespace (window style). There are multiple reasons why Twine should allow to register modules, constructors and/or functions.

This PR adds a public method `Twine.register(name, constructor)`, this allows us to register constructors with Twine without having to attach them to the global namespace, e.g. `window.MyModule`

#### With classes

```coffee
class MyModule
  # ...

Twine.register('MyModule', MyModule)
```

```html
<div data-define="{myModule: new MyModule()}">
```

#### With object literals

```coffee
Twine.register 'MyModule', {
  fn: ->
    # ...
}
```

```html
<div data-define="{myModule: MyModule.fn()}">
```

#### With functions

```coffee
Twine.register 'MyFn', ->
  # ...
```

```html
<div data-define="{myModule: MyFn()}">
```

@Shopify/tnt


lib/assets/javascripts/twine.js.coffee
```diff
@@ -258,47 +267,47 @@ Twine.bindingTypes =
...
     return refresh: ->
-      newValue = fn.call(node, context, rootContext)
+      newValue = fn.call(node, context, rootContext, modules)
       for key, value of newValue when !lastValue[key] != !value
         $(node).toggleClass(key, !!value)
       lastValue = newValue
 
   'bind-attribute': (node, context, definition) ->
-    fn = wrapFunctionString(definition, '$context,$root', node)
+    fn = wrapFunctionString(definition, '$context,$root,$modules', node)
```
DrewMartin: Not sure if we would need/want modules to be available outside of define/eval. Hmmm

tsov:
seem legit 🔴  will fix soon

lib/assets/javascripts/twine.js.coffee
tsov: Good point, you'd ideally add your registered "component" to the context through `define` right?

DrewMartin: Or eval. I can't think of a situation off the top of my head where we'd need it on other things but it might not hurt to also include it there. Hmm hmm hmm

tsov: @Shopify/tnt lets vote maybe :P ?

- 👍  yes
- 👎 no

qq99: I would say if there isn't a good reason to expose it, don't

dan-menard: > I would say if there isn't a good reason to expose it, don't

This very yes. Easy to add more exposure later if needed, hard to take exposure away if it turns out to be a bad idea.

lib/assets/javascripts/twine.js.coffee
```diff
@@ -194,7 +203,7 @@ wrapFunctionString = (code, args, node) ->
       ($context, $root) -> getValue($context, keypath)
   else
     try
-      new Function(args, "with($context) { return #{code} }")
+      new Function(args, "with($context) { with($modules) { return #{code} } }")
```
tsov: Rather than changing the function string above to do some conditionals it was easier to pass in an empty object literal for `$modules` for all but `define` & `eval`. Seemed like the nicer fix, any other suggestions are more than welcome 👍

pushrax: Can you prove this doesn't affect performance with jsperf?

tsov: screen shot 2016-06-17 at 1 59 55 pm

gif-keyboard-3342855315800191201

dist/twine.js
```diff
@@ -144,6 +146,14 @@ Twine.refreshImmediately = function() {
   }
 };
 
+Twine.register = function(name, constructor) {
```
lemonmade: Can you use a different name? `constructor` isn't a reserved word but it's probably better to stay away from it anyways.

tsov: for sure! Initially I figured we might only accept constructors, what about `module` ?

dist/twine.js
```diff
@@ -8,6 +8,8 @@ Twine.shouldDiscardEvent = {};
 
 elements = {};
 
+modules = {};
```
tsov: What do people think about naming this `registry`/`moduleRegistry` instead?

dist/twine.js
lemonmade: `component` maybe?

dist/twine.js
```diff
@@ -314,7 +324,7 @@ wrapFunctionString = function(code, args, node) {
     }
   } else {
     try {
-      return new Function(args, "with($context) { return " + code + " }");
+      return new Function(args, "with($context) { with($modules) { return " + code + " } }");
```
lemonmade: @dan-menard rejoice! nested `with`!

lib/assets/javascripts/twine.js.coffee
lemonmade: Derp, just realized I was commenting on the list version :(

tsov: Ok, did some benchmarks and read some articles:

### Performance pro & contra

__Pro:__ The `with` statement can help reduce file size by reducing the need to repeat a lengthy object reference without performance penalty. The scope chain change required by 'with' _is not computationally expensive_. Use of `with` will _relieve the interpreter_ of parsing repeated object references. Note, however, that in many cases this benefit can be achieved by using a temporary variable to store a reference to the desired object.

__Contra:__ The with statement forces the specified object to be searched first for all name lookups. Therefore all identifiers that aren't members of the specified object will be found more slowly in a `with` block. Where performance is important, `with` should only be used to encompass code blocks that access members of the specified object.

----

> `with` should only be used to encompass code blocks that access members of the specified object.

I think this is the biggest take away ^ So we should use `with` conditionally, SO when using `define` or `eval` it wraps the nested `with` which includes `$modules` otherwise it remains as is

Something like?

```coffee
"""
with ($context) {
  if (typeof $modules != 'undefined') {
    with ($modules) {
      return #{code}
    }
  } else {
    return #{code}
  }
}
"""
```

Any thoughts?

pushrax: Why not simply generate a different function to begin with?

lib/assets/javascripts/twine.js.coffee
```diff
@@ -193,11 +202,24 @@ wrapFunctionString = (code, args, node) ->
...
+expressionFactory = (args, code) ->
+  if requiresRegistry(args)
+    wrap.registryFn(code)
+  else
+    wrap.defaultFn(code)
+
+wrap =
+  defaultFn: (code) -> "with($context) { return #{code} }"
+  registryFn: (code) -> "with($context) { with($registry) { return #{code} } }"
+
```
tsov: @DrewMartin You can add the array iterator function to this object here ^

tsov:
Updated, can I get some 👀  again?

DrewMartin:
:+1:

asmockler:
LGTM

Issue #73, opened 2016-06-14
Proposal: array bindings


DrewMartin:
### Problem
Right now we have an awkward way of setting up arrays of objects inside of twine. For instance, if you're on a product page and you want to initialize a list of product variant objects, each with the same keys and different values you've got a few options:

#### Option 1: make the define names unique
When defining each of them you can add some unique key, like the variant id, to the name:
`define="{variant<% variant.id %>: new ProductVariant(...)}"` (optionally `context="variant<% variant.id %>"`)
This isn't great because now you can't loop over the objects easily inside the parent object, and it's pretty ugly because you need to use the generated key everywhere.

#### Option 2: overwrite each object successively and use contexts to access them
Each iteration overwrites the previous object, but keeps it in a closure via `context`:
`define="{variant: new ProductVariant(...)}" context="variant"`
Inside the element with the context everything is bound to the correct variant, but the parent object can't access the different variants except for the last one.

#### Option 3: eval
You can use the slightly hacky `eval` to initialize the array and push to it:
```html
<div define="{variants: []}">
  <% variants.each do |variant| %>
    <div eval="variants.push(new ProductVariant(...));">
```
This works but isn't the most friendly looking. `eval` is basically define, in that it will run arbitrary code, but it does nothing with the return value.

### Proposal
I propose an array syntax for use inside define:
`define="{variants[]: new ProductVariant(...)}"`
when the last two characters of a key to define are `[]` it will create an array with that name if it doesn't exist, or push the object to the end of the array (unless it exists and it's not an array, in which case raise). You can do `context="variants[]"` and it will set the context to be the last thing in the array. 

The biggest downside I see is that it makes it difficult to access the right variant from the array from inside the parent context, which only option 1 from above lets you do. If needed you could definitely do something like this:
```html
<% variants.each_with_index do |variant, i| %>
  <div define="{variants[]: new ProductVariant(...)}">
    <input bind="variants[<% i %>].price">
```

We might be able to have some magic that keeps the correct index around and used implicitly so that you could instead do `<input bind="variants[].price">` and have it point to the correct variant, but I'd have to play around with it. It would also potentially become a mess if things started getting added and removed. Like if you did a partial refresh detecting that all the variants are being replaced and so the array should empty out might be tricky, but maybe that's the developer's responsibility instead of twine's.

Thoughts? Alternatives? Criticisms?
@Shopify/tnt

qq99:
I worry most about partial replacing & when to clear the array, when to replace elements

[...]

PR #76, opened 2016-06-20
define-array and using the array index implicitly

DrewMartin:
Implementation of https://github.com/Shopify/twine/issues/73

This lets you use `define-array` (or `data-define-array`) similar to define, except it stores the returned values in an array and will push new values to the end. We also store some metadata about array indexes behind the scenes to know which elements correspond to which iterations of an array. Every element (that has bindings) have the same array indexes of its parent, with any new indexes as well. When a new context is entered the nodes inside of it start with a fresh set of array indexes as well, unable to see the array indexes on nodes outside of the current context.
For example:
```html
<div data-define-array="{key: {value: 5, foo: function(){ return 12; }}}">
  <div data-define-array="{nested: {value: 50}}">
    <span bind="key.value">5</span> <!-- key[0].value -->
    <span bind="key.value">12</span> <!-- key[0].foo() -->
    <span bind="nested.value">50</span> <!-- nested[0].value -->
  </div>
  <div data-define-array="{nested: {value: 75}}">
    <span bind="key.value">5</span> <!-- key[0].value -->
    <span bind="key.value">12</span> <!-- key[0].foo() -->
    <span bind="nested.value">75</span> <!-- nested[1].value -->
  </div>
</div>
<div data-define-array="{key: {value: 10, foo: function(){ return 99; }}}">
  <div data-define-array="{nested: {value: 100}}">
    <span bind="key.value">10</span> <!-- key[1].value -->
    <span bind="key.value">99</span> <!-- key[1].foo() -->
    <span bind="nested.value">100</span> <!-- nested[2].value -->
  </div>
  <div data-define-array="{nested: {value: 125}}">
    <span bind="key.value">10</span> <!-- key[1].value -->
    <span bind="key.value">99</span> <!-- key[1].foo() -->
    <span bind="nested.value">125</span> <!-- nested[3].value -->
  </div>
</div>
<div data-define-array="{key: {value: 42}}" context="key"> <!-- key[2] -->
  <!-- has no indexes available -->
  <span bind="value">42</span>
</div>
```

This takes two main changes:
* when building the keypath for a binding, which is used for getting and setting simple values, where it would be `["nested", "value"]` for `nested.value`, instead it becomes `["nested", 0, "value"]` (or whatever the correct array index is, pulled from the node itself). This will only happen for the first key in the keypath because that's the potential array on the context and everything else is nested within the context and not created with `define-array`, at least not on the element being evaluated. (you could have a `define-array` -> `context` -> `define-array` but the outer node wouldn't have access to the array indexes on the inner `define-array`)
* when the binding does not resolve to a keypath, so basically when it's a function call, we create a hash of objects inside the array(s) and use it as the innermost `with` expression, inside of the `with($context)`, so that these take precedence. Effectively it does something like this:
```js
with({key: [1, 2, 3]}) {
  with({key: 1}) { // Or {key: 2}, or {key: 3} depending on what the index is for this node in the array `key` 
    return // Code here
  }
}
```

@pushrax @Shopify/tnt

### DIFF
--- a/dist/twine.js
+++ b/dist/twine.js
@@ -62,17 +62,19 @@
       }
     };
     bind = function(context, node, indexes, forceSaveContext) {
-      var binding, callback, callbacks, childNode, defineArrayAttr, definition, element, fn, j, k, key, keypath, len, len1, newContextKey, newIndexes, ref, ref1, ref2, ref3, type, value;
+      var binding, callback, callbacks, childNode, defineArrayAttr, definition, element, fn, j, k, key, keypath, len, len1, newContextKey, newIndexes, ref, ref1, ref2, type, value;
       currentBindingCallbacks = [];
       if (node.bindingId) {
         Twine.unbind(node);
       }
       if (defineArrayAttr = Twine.getAttribute(node, 'define-array')) {
         newIndexes = defineArray(node, context, defineArrayAttr);
-        ref = indexes || {};
-        for (key in ref) {
-          value = ref[key];
-          if (newIndexes[key] == null) {
+        if (indexes == null) {
+          indexes = {};
+        }
+        for (key in indexes) {
+          value = indexes[key];
+          if (!newIndexes.hasOwnProperty(key)) {
             newIndexes[key] = value;
           }
         }
@@ -80,9 +82,9 @@
         element = findOrCreateElementForNode(node);
         element.indexes = indexes;
       }
-      ref1 = Twine.bindingTypes;
-      for (type in ref1) {
-        binding = ref1[type];
+      ref = Twine.bindingTypes;
+      for (type in ref) {
+        binding = ref[type];
         if (!(definition = Twine.getAttribute(node, type))) {
           continue;
         }
@@ -109,21 +111,22 @@
       if (element || newContextKey || forceSaveContext) {
         element = findOrCreateElementForNode(node);
         element.childContext = context;
-        if (element.indexes == null) {
-          element.indexes = indexes;
+        if (indexes != null) {
+          if (element.indexes == null) {
+            element.indexes = indexes;
+          }
         }
-        elements[node.bindingId != null ? node.bindingId : node.bindingId = ++nodeCount] = element;
       }
       callbacks = currentBindingCallbacks;
-      ref2 = node.children || [];
-      for (j = 0, len = ref2.length; j < len; j++) {
-        childNode = ref2[j];
+      ref1 = node.children || [];
+      for (j = 0, len = ref1.length; j < len; j++) {
+        childNode = ref1[j];
         bind(context, childNode, newContextKey != null ? null : indexes);
       }
       Twine.count = nodeCount;
-      ref3 = callbacks || [];
-      for (k = 0, len1 = ref3.length; k < len1; k++) {
-        callback = ref3[k];
+      ref2 = callbacks || [];
+      for (k = 0, len1 = ref2.length; k < len1; k++) {
+        callback = ref2[k];
         callback();
       }
       currentBindingCallbacks = null;
--- a/dist/twine.js
+++ b/dist/twine.js
@@ -10,7 +10,7 @@
       return root.Twine = factory();
     }
   })(this, function() {
-    var Twine, attribute, bind, currentBindingCallbacks, elements, eventName, fireCustomChangeEvent, getContext, getValue, isKeypath, j, k, keypathForKey, keypathRegex, len, len1, nodeCount, preventDefaultForEvent, ref, ref1, refreshElement, refreshQueued, registry, requiresRegistry, rootContext, rootNode, setValue, setupAttributeBinding, setupEventBinding, stringifyNodeAttributes, valueAttributeForNode, wrapFunctionString;
+    var Twine, arrayPointersForNode, attribute, bind, currentBindingCallbacks, defineArray, elements, eventName, findOrCreateElementForNode, fireCustomChangeEvent, getContext, getIndexesForElement, getValue, isKeypath, j, k, keyWithArrayIndex, keypathForKey, keypathRegex, len, len1, nodeCount, nodeHasArrayIndexes, preventDefaultForEvent, ref, ref1, refreshElement, refreshQueued, registry, requiresRegistry, rootContext, rootNode, setValue, setupAttributeBinding, setupEventBinding, stringifyNodeAttributes, valueAttributeForNode, wrapFunctionString;
     Twine = {};
     Twine.shouldDiscardEvent = {};
     elements = {};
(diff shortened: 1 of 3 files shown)


lib/assets/javascripts/twine.coffee
```diff
@@ -147,6 +166,13 @@
         return context
       node = node.parentNode if child
 
+  getIndexesForElement = (node) ->
+    firstContext = null
+    while node
+      return elements[id]?.indexes if id = node.bindingId
```
DrewMartin: Since we assign the same indexes to all child nodes that have any bindings, if a node has a bindingId it will also have the set of indexes to use, if any.

lib/assets/javascripts/twine.coffee
```diff
@@ -46,33 +46,47 @@
...
 
-  bind = (context, node, forceSaveContext) ->
+  bind = (context, node, indexes, forceSaveContext) ->
     currentBindingCallbacks = []
     if node.bindingId
       Twine.unbind(node)
 
+    if defineArrayAttr = Twine.getAttribute(node, 'define-array')
+      newIndexes = defineArray(node, context, defineArrayAttr)
+      for key, value of indexes || {} when !newIndexes[key]?
```
dan-menard: Didn't know you could use `when` like that 👍

lib/assets/javascripts/twine.coffee
```diff
@@ -46,33 +46,47 @@
...
+      indexes = newIndexes
+      # register the element early because subsequent bindings on the same node might need to make use of the index
+      element = findOrCreateElementForNode(node)
+      element.indexes = indexes
+
     for type, binding of Twine.bindingTypes when definition = Twine.getAttribute(node, type)
-      element = {bindings: []} unless element  # Defer allocation to prevent GC pressure
+      element = findOrCreateElementForNode(node)
+      element.bindings ?= []
+      element.indexes ?= indexes
```
dan-menard: You're sure you can use `?=` here? That checks explicitly for `null` and nothing else. If `element.indexes` is `undefined`, it will stay `undefined` in this case. Is it maybe safer to use `element.indexes ||= indexes` and cover both cases?

lib/assets/javascripts/twine.coffee
```diff
@@ -147,6 +166,13 @@
         return context
       node = node.parentNode if child
 
+  getIndexesForElement = (node) ->
+    firstContext = null
+    while node
+      return elements[id]?.indexes if id = node.bindingId
+      node = node.parentNode
+    return
```
dan-menard: No need for this.

lib/assets/javascripts/twine.coffee
```diff
@@ -219,6 +259,20 @@
...
   requiresRegistry = (args) -> /\$registry/.test(args)
 
+  nodeHasArrayIndexes = (node) ->
+    return unless node.bindingId?
+    !!elements[node.bindingId]?.indexes?
+
+  arrayPointersForNode = (node, context) ->
+    return {} unless node.bindingId?
+    indexes = elements[node.bindingId]?.indexes
+    return {} unless !!indexes?
```
dan-menard: This is clever, but a pretty unclear. What do you think of making the "allow 0" aspect a little more explicit?

`return {} unless indexes || indexes == 0`

dan-menard:
A few code comments, but the syntax and overall implementation look solid to me 👍

lib/assets/javascripts/twine.coffee
DrewMartin: I don't know why I have !!, `indexes?` alone would do it

lib/assets/javascripts/twine.coffee
DrewMartin: It checks for `== null` in javascript, and `undefined == null` is true. The `?` operator in cs checks if something is undefined or null

dan-menard: Ah! Yeah, sorry, mistook the == for === when I ran that through a to-JS converter. Code's good as-is 👍

lib/assets/javascripts/twine.coffee
dan-menard: Oh, true! Yeah, I'm fine with just `indexes?`

dist/twine.js
```diff
@@ -61,55 +61,84 @@
...
       }
-      ref = Twine.bindingTypes;
-      for (type in ref) {
-        binding = ref[type];
+      if (defineArrayAttr = Twine.getAttribute(node, 'define-array')) {
+        newIndexes = defineArray(node, context, defineArrayAttr);
+        ref = indexes || {};
+        for (key in ref) {
+          value = ref[key];
+          if (newIndexes[key] == null) {
```
lemonmade: `hasOwnProperty` is probably more what you are after here.

lib/assets/javascripts/twine.coffee
lemonmade: Can you pull the `indexes || {}` part up to a dedicated variable? This line is really hard to parse. Also, `hasOwnProperty` is probably more what you are after for the `newIndexes[key]` part.

qq99: `indexes = {}` default in the method signature?

lib/assets/javascripts/twine.coffee
```diff
@@ -46,33 +46,47 @@
...
+      newIndexes = defineArray(node, context, defineArrayAttr)
+      for key, value of indexes || {} when !newIndexes[key]?
+        newIndexes[key] = value
+      indexes = newIndexes
+      # register the element early because subsequent bindings on the same node might need to make use of the index
+      element = findOrCreateElementForNode(node)
+      element.indexes = indexes
+
     for type, binding of Twine.bindingTypes when definition = Twine.getAttribute(node, type)
-      element = {bindings: []} unless element  # Defer allocation to prevent GC pressure
```
qq99: how did this work? I don't get the `unless element` on same line

DrewMartin: Since this is in a loop the first iteration will initialize it. It looks like it wasn't being initialized outside because of the large amount of elements in the dom that wouldn't have any bindings, so that we didn't create and then throw away the `element` object, making the GC work harder

lib/assets/javascripts/twine.coffee
DrewMartin: I don't want to do indexes as a default arg because that would mean a lot of elements storing an empty object as the indexes, which means we have to check the number of keys in an object to tell if an element has indexes rather than just a null check.

Better to keep it as null/undefined when there are no indexes to store.

DrewMartin: (It would also affect the assignment on line 89)

DrewMartin:
Everyone cool with this? 🚢 ?

tsov:
👍  we can iterate on it too.

Would you mind adding a commit with README docs for `define-array` to https://github.com/Shopify/twine/pull/77 ?

DrewMartin:
It doesn't look like it really belongs in the readme so much as http://shopify.github.io/twine/ or the docs @dan-menard's working on, but I'll get it added to one or both

PR #77, opened 2016-06-21
Bump to 1.0.0

tsov:
Let's bump up Twine version to 1.0.0 and start using it through npm instead.

We should now start using `require('twine')` or `import Twine from 'twine'`.

@Shopify/tnt @Shopify/admin-fed

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,9 @@ Twine is available on bower via `bower install twine` if that is your preference
 
 Twine comes as `dist/twine.js` and `dist/twine.min.js` in this repo and in the bower package.
 
-Twine is also available as a gem.  In your Gemfile, add `gem 'twine-rails'` and include it in your `application.js` manifest via `//= require twine`
+Twine is also available as a gem. In your Gemfile, add `gem 'twine-rails'` and include it in your `application.js` manifest via `//= require twine`
+
+AMD, CommonJS and Browser global (using UMD) are also supported.
 
 ## Usage
 
@@ -127,6 +129,24 @@ Example:
   Twine.shouldDiscardEvent.click = (event) ->
     $target = $(event.target)
     $target.hasClass('disabled')
+```
+
+## Twine.register
+
+Lets you add constructors, modules, functions, etc to Twine that are not globally available. This means you can keep your classes etc
+as local variables and Twine will find them for you within `define`s & `eval`s.
+
+```coffee
+  # local_class.coffee
+
+  class LocalClass
+    # ...
+
+  Twine.register('LocalClass', LocalClass)
+```
+
+```html
+  <div define="{localClass: new LocalClass()}"></div>
 ```
 
 ## Dev Console
--- a/package.json
+++ b/package.json
@@ -1,11 +1,14 @@
 {
   "name": "twine",
-  "version": "0.1.7",
+  "version": "1.0.0",
   "description": "A minimalistic 2-way binding system",
-  "main": "index.js",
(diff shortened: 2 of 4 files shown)


lemonmade:
Can we point the `package.json` `main` field at `twine/dist/twine` so that we can just require the package itself?

qq99:
How about 1.0.0?

tsov:
@lemonmade will do! just talking to @bouk about publishing twine @npm, but `twine` package name is taken.

@qq99 I wouldn't say it's major changes, also the last changes are all backwards compatible, maybe with @DrewMartins work

DrewMartin:
Yeah if we're going to bump major or minor version number can we do it after https://github.com/Shopify/twine/pull/76?

I'm not sure if that's needed either way. Everything is backwards compatible so only a patch version change is really needed.

lemonmade:
We have been using this for several years, 1.0 probably makes sense. Semver < 1.0 also has a different meaning than >= 1.0, so it would be nice to get to a place where people understand the meaning of version number changes.

bouk:
Would there be a way to share the version name between ruby and the package.json without duplicating it? Maybe the ruby file could read the package.json?

Also does anyone actually use the bower package?

tsov:
@DrewMartin why don't you just bump the minor version again after your PR? I increased the minor version because there is added functionality, which is backwards compatible. If we increased major version it would indicate that there are breaking changes.

@bouk was thinking the same, but the bower package kinda ruins this for us haha

DrewMartin:
So go from version 0.2 to 0.3 within a couple days? Unless we need to get a new release in immediately or #76 will take weeks to review I'd say just wait

tsov:
> So go from version 0.2 to 0.3 within a couple days? 

I don't see a problem with that, versions should not represent time, they represent changes. I don't mind either way

bouk:
I think you should release this onto npm as version 1.0, when you get an appropriate name for it

tsov:
Sweet, we now have `twine` package name on npm 👌 

Let's use this PR to get twine ready for publishing on npm, e.g. version bump to 1.0.0, appropriate changes to `package.json`

Then once https://github.com/Shopify/twine/pull/76 is merged we can ship this PR and publish on npm

package.json
```diff
@@ -1,8 +1,8 @@
 {
```
bouk: you should add an `.npmignore` that excludes basically everything but the `dist` folder so we don't include stuff like ruby files in the npm package

tsov: We could also use the `files` field in `package.json` to whitelist the `dist` folder and other files, e.g. readme?

bouk: yeah that's even better :+1:

DrewMartin:
This should be good to merge

tsov:
@DrewMartin just changing `0.2.0` to `1.0.0` and adding the npm whitelist, then 🚢 ing

tsov:
@lemonmade @bouk all good to go?

lemonmade:
👍

README.md
```diff
@@ -129,6 +131,24 @@ Example:
...
+  class LocalClass
+    # ...
+
+  Twine.register('LocalClass', LocalClass)
+```
+
+```html
+  <div define="{localClass: new LocalClass()}"></div>
+```
+
```
bouk: Neat

bouk:
Lgtm

tsov:
fyi: https://www.npmjs.com/package/twine

PR #78, opened 2016-06-23
Document define-array changes

DrewMartin:
I noticed `context` wasn't documented in here and I wanted to show how to use it with `define-array` so I added documentation about using it with `define` as well

@Shopify/tnt


index.html
```diff
@@ -193,6 +193,59 @@
 
 <p>This can be pretty useful for when you want to use <code>bind-show</code> and start in a visible state.</p>
 
+<p><code>context</code> lets you change the current context for every element nested inside.</p>
+
+<div class="prepend-pre">
+<div define="{object: {value: 'foo'} }" context="foo">
```
tylerball: Wouldn't this be `context="object"`? Or I guess `{foo: {value: 'foo'}}`

DrewMartin: Yep!

qq99:
I think you might also need to merge in the js? Since this page should be eval&running it, I'm pretty sure

index.html
```diff
@@ -193,6 +193,59 @@
...
+<div>
+  <span bind="foo.value"> <!-- foo.value from outside the context of foo -->
+</div>
+</div>
+
+<h2>Define-array</h2>
+<p><code>define-array</code> lets you define an array of objects one at a time in the context and implicitly reference the correct member in the array.</p>
+
+<div class="prepend-pre">
+<div data-define-array="{key: {value: 5, foo: function(){ return 12; }}}">
```
tylerball: The name `key` is kind of confusing here. It made me think that it was a keyword with some kind of special behaviour related to `define-array` when really its just the name of the array you're appending to. I prefer names like `widgets` or `things` in docs like this.

index.html
```diff
@@ -193,6 +193,59 @@
...
+  <div data-define-array="{nested: {value: 50}}">
```
tylerball: rebasing off of master, `nested` is still a top-level context, so is `nested` a good name for this?

DrewMartin: Yeah the purpose of that was to show that you could have an array defined inside the definition of another array and things would still work, without changing the context

index.html
DrewMartin: Yeah true I'll update it

index.html
tylerball: Ah I get it.

DrewMartin:
I merged in master. dat commit size though. The only file changed in here is still `index.html`. Updated it to use nicer keys

tylerball:
Nice 👍

PR #79, opened 2016-07-05
Bind performance

justinthec:
### Why?
After some investigation into the performance of `Twine.bind` in [this doc](https://docs.google.com/document/d/1tA2zFtiY-J0PcYHqQkw59Ma6mK4pgWnXfBCTRowZ-s8/), @bouk found that the iteration over every binding type in `Twine.bindingTypes` was unnecessary and it would be more performant to instead iterate over the attributes of the node.

### Performance Test
I did a performance test on Shopify core by writing some equivalent JS on the master branch:
```js
element = findOrCreateElementForNode(node);
if (element.bindings == null) {
  element.bindings = [];
}
if (element.indexes == null) {
  element.indexes = indexes;
}

Array.from(node.attributes).forEach((attribute) => {
  let type = attribute.name
  let definition = attribute.value
  if (type.startsWith('data-')) {
    type = type.slice(5);
  }

  binding = Twine.bindingTypes[type]
  if (!binding) {
    return;
  }
  fn = binding(node, context, definition, element);
  if (fn) {
    element.bindings.push(fn);
  }
});
```

js <-- coffee by js2.coffee output:
image

On the Home page the duration of `Twine.bind` caused by the page load was recorded as follows:

| Trial     | Before    | After    |
|-----------|-----------|----------|
| 1         | 148.07ms  | 129.85ms |
| 2         | 155.89ms  | 112.52ms |
| 3         | 170.32ms  | 123.46ms |
| 4         | 155.71ms  | 150.73ms |
| 5         | 168.24ms  | 118.74ms |
|           |           |          |
| Averages: | **159.646ms** | **127.06ms** |

### Changes
 - The `bind` function now iterates over `node.attributes` rather than `Twine.bindingTypes` since `node.attributes` will most likely be the smaller set.
 - Moved the `element = findOrCreateElementForNode` assignment outside of the loop to prevent unnecessary calls and repeated assignments.

**For review:**
@bouk @lemonmade @Shopify/tnt
cc
@Shopify/admin-fed @edward @GoodForOneFare

### DIFF
--- a/lib/assets/javascripts/twine.coffee
+++ b/lib/assets/javascripts/twine.coffee
@@ -56,6 +56,7 @@
 
   bind = (context, node, indexes, forceSaveContext) ->
     currentBindingCallbacks = []
+    element = null
     if node.bindingId
       Twine.unbind(node)
 
@@ -69,18 +70,29 @@
       element = findOrCreateElementForNode(node)
       element.indexes = indexes
 
-    element = findOrCreateElementForNode(node)
-    element.bindings ?= []
-    element.indexes ?= indexes
-    Array.prototype.slice.call(node.attributes).forEach (attribute) ->
+    bindingConstructors = null
+    for attribute in node.attributes
       type = attribute.name
       type = type.slice(5) if type.slice(0, 5) == 'data-'
       definition = attribute.value
 
-      binding = Twine.bindingTypes[type]
-      return unless binding
-      fn = binding(node, context, definition, element)
-      element.bindings.push(fn) if fn
+      constructor = Twine.bindingTypes[type]
+      continue unless constructor
+
+      bindingConstructors ?= []
+      if type == 'bind'
+        bindingConstructors.unshift({constructor, definition})
+      else
+        bindingConstructors.push({constructor, definition})
+
+    if bindingConstructors
+      element ?= findOrCreateElementForNode(node)
+      element.bindings ?= []
+      element.indexes ?= indexes
+
+      for {constructor, definition} in bindingConstructors
+        binding = constructor(node, context, definition, element)
+        element.bindings.push(binding) if binding
 
     if newContextKey = Twine.getAttribute(node, 'context')
       keypath = keypathForKey(node, newContextKey)
@@ -90,7 +102,7 @@
       context = getValue(context, keypath) || setValue(context, keypath, {})
 
     if element || newContextKey || forceSaveContext
-      element = findOrCreateElementForNode(node)
+      element ?= findOrCreateElementForNode(node)
       element.childContext = context
       element.indexes ?= indexes if indexes?
 
--- a/lib/assets/javascripts/twine.coffee
+++ b/lib/assets/javascripts/twine.coffee
@@ -72,7 +72,7 @@
     element = findOrCreateElementForNode(node)
     element.bindings ?= []
     element.indexes ?= indexes
-    Array.from(node.attributes).forEach (attribute) ->
+    Array.prototype.slice.call(node.attributes).forEach (attribute) ->
       type = attribute.name
       type = type.slice(5) if type.startsWith('data-')
       definition = attribute.value
--- a/lib/assets/javascripts/twine.coffee
+++ b/lib/assets/javascripts/twine.coffee
@@ -283,7 +283,11 @@
     value not in ['true', 'false', 'null', 'undefined'] && keypathRegex.test(value)
 
   isDataAttribute = (value) ->
-    value[0] == 'd' && value[1] == 'a' && value[2] == 't' && value[3] == 'a' && value[4] == '-'
+    value[0] == 'd' &&
+    value[1] == 'a' &&
+    value[2] == 't' &&
+    value[3] == 'a' &&
+    value[4] == '-'
 
   fireCustomChangeEvent = (node) ->
     event = document.createEvent('CustomEvent')
--- a/lib/assets/javascripts/twine.coffee
+++ b/lib/assets/javascripts/twine.coffee
@@ -69,11 +69,16 @@
       element = findOrCreateElementForNode(node)
       element.indexes = indexes
 
-    for type, binding of Twine.bindingTypes when definition = Twine.getAttribute(node, type)
-      element = findOrCreateElementForNode(node)
-      element.bindings ?= []
-      element.indexes ?= indexes
+    element = findOrCreateElementForNode(node)
+    element.bindings ?= []
+    element.indexes ?= indexes
+    Array.from(node.attributes).forEach (attribute) ->
+      type = attribute.name
+      type = type.slice(5) if type.startsWith('data-')
+      definition = attribute.value
 
+      binding = Twine.bindingTypes[type]
+      return unless binding
       fn = binding(node, context, definition, element)
       element.bindings.push(fn) if fn
 
--- a/lib/assets/javascripts/twine.coffee
+++ b/lib/assets/javascripts/twine.coffee
@@ -381,7 +381,7 @@
   setupPropertyBinding = (attributeName, bindingName) ->
     booleanProp = attributeName in ['checked', 'indeterminate', 'disabled', 'readOnly']
 
-    Twine.bindingTypes["bind-#{bindingName}"] = (node, context, definition) ->
+    Twine.bindingTypes["bind-#{bindingName.toLowerCase()}"] = (node, context, definition) ->
       fn = wrapFunctionString(definition, '$context,$root,$arrayPointers', node)
       lastValue = undefined
       return refresh: ->
--- a/lib/assets/javascripts/twine.coffee
+++ b/lib/assets/javascripts/twine.coffee
@@ -73,24 +73,24 @@
     bindingConstructors = null
     for attribute in node.attributes
       type = attribute.name
-      type = type.slice(5) if type.slice(0, 5) == 'data-'
-      definition = attribute.value
+      type = type.slice(5) if isDataAttribute(type)
 
       constructor = Twine.bindingTypes[type]
       continue unless constructor
 
       bindingConstructors ?= []
+      definition = attribute.value
       if type == 'bind'
-        bindingConstructors.unshift({constructor, definition})
+        bindingConstructors.unshift([constructor, definition])
       else
-        bindingConstructors.push({constructor, definition})
+        bindingConstructors.push([constructor, definition])
 
     if bindingConstructors
       element ?= findOrCreateElementForNode(node)
       element.bindings ?= []
       element.indexes ?= indexes
 
-      for {constructor, definition} in bindingConstructors
+      for [constructor, definition] in bindingConstructors
         binding = constructor(node, context, definition, element)
         element.bindings.push(binding) if binding
 
@@ -281,6 +281,9 @@
 
   isKeypath = (value) ->
     value not in ['true', 'false', 'null', 'undefined'] && keypathRegex.test(value)
+
+  isDataAttribute = (value) ->
+    value[0] == 'd' && value[1] == 'a' && value[2] == 't' && value[3] == 'a' && value[4] == '-'
 
   fireCustomChangeEvent = (node) ->
     event = document.createEvent('CustomEvent')
--- a/lib/assets/javascripts/twine.coffee
+++ b/lib/assets/javascripts/twine.coffee
@@ -74,7 +74,7 @@
     element.indexes ?= indexes
     Array.prototype.slice.call(node.attributes).forEach (attribute) ->
       type = attribute.name
-      type = type.slice(5) if type.startsWith('data-')
+      type = type.slice(5) if type.slice(0, 5) == 'data-'
       definition = attribute.value
 
       binding = Twine.bindingTypes[type]
--- a/test/twine_test.coffee
+++ b/test/twine_test.coffee
@@ -164,28 +164,53 @@ suite "Twine", ->
       node = setupView(testView, key: "<script>")
       assert.equal node.innerHTML, "&lt;script&gt;"
 
+    test "should be the first binding to run on input event", ->
+      testView = "<input type=\"text\" bind-event-input=\"eventFunc()\" data-bind=\"val\">"
+      context = {
+        eventFunc: @spy(),
+      }
+      bindGetter = @spy()
+      Object.defineProperty(context, 'val', {
+        get: bindGetter
+      })
+      node = setupView(testView, context)
+      context.eventFunc.reset()
+      bindGetter.reset()
+
+      triggerEvent node, "input"
+      sinon.assert.callOrder bindGetter, context.eventFunc
+
+    test "should be the first binding to run on keyup event", ->
+      testView = "<input type=\"text\" bind-event-keyup=\"eventFunc()\" data-bind=\"val\">"
+      context = {
+        eventFunc: @spy(),
+      }
+      bindGetter = @spy()
+      Object.defineProperty(context, 'val', {
+        get: bindGetter
+      })
+      node = setupView(testView, context)
+      context.eventFunc.reset()
+      bindGetter.reset()
+
+      triggerEvent node, "keyup"
+      sinon.assert.callOrder bindGetter, context.eventFunc
+
     test "should be the first binding to run on change event", ->
       testView = "<input type=\"text\" bind-event-change=\"eventFunc()\" data-bind=\"val\">"
       context = {
-        val: 1,
-        eventFunc: () -> this.val = this.val * 2
+        eventFunc: @spy(),
       }
+      bindGetter = @spy()
+      Object.defineProperty(context, 'val', {
+        get: bindGetter
+      })
       node = setupView(testView, context)
-      node.value = 2
+      context.eventFunc.reset()
+      bindGetter.reset()
+
       triggerEvent node, "change"
-
-      assert.equal context.val, 4
-
-    # test "should be the first binding to run on input event", ->
-    #   testView = "<input type=\"text\" bind-event-input=\"eventFunc()\" data-bind=\"bindFunc()\">"
-    #   node = setupView(testView, context = {
-    #     eventFunc: @spy(),
-    #     bindFunc: @spy()
-    #   })
-    #   context.eventFunc.reset()
-    #   context.bindFunc.reset()
-    #   triggerEvent node, "input"
-    #   sinon.assert.callOrder context.bindFunc, context.eventFunc
+      sinon.assert.callOrder bindGetter, context.eventFunc
 
   suite "data-bind-show attribute", ->
     test "should apply the \"hide\" class when falsy", ->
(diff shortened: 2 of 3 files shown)


DrewMartin:
See https://github.com/Shopify/twine/pull/31 and https://github.com/Shopify/twine/pull/33

lib/assets/javascripts/twine.coffee
```diff
@@ -69,11 +69,16 @@
       element = findOrCreateElementForNode(node)
       element.indexes = indexes
 
-    for type, binding of Twine.bindingTypes when definition = Twine.getAttribute(node, type)
-      element = findOrCreateElementForNode(node)
-      element.bindings ?= []
-      element.indexes ?= indexes
-
+    element = findOrCreateElementForNode(node)
```
DrewMartin: Also this will cause twine to take up way more memory than we currently do because we'd store bindings for every node in the DOM, rather than just those with bindings as we do now

justinthec:
@DrewMartin 
From #33:
>We could theoretically do something like try to handle bind before other attributes, but this would lead to code duplicat[ion]

If this endeavour could bring potential performance gains of reducing `Twine.bind` time by 70% across all pages of the admin, it might be worth doing some code duplication considering Shopify is the primary client of Twine.

Elaborating on what it would take to handle `bind` before other `bindings`, my understanding is that it would require a change to the object that [`Twine.bindingTypes`'s `bind` function](https://github.com/Shopify/twine/blob/master/lib/assets/javascripts/twine.coffee#L304) returns to include some sort of key to identify this binding as the `bind` binding. The [`refreshElement`](https://github.com/Shopify/twine/blob/master/lib/assets/javascripts/twine.coffee#L120) function would then have to search the `bindings` list for the object with the `bind` flag and `refresh` that binding first, then loop over the rest of them.

Correct me if I'm wrong?

DrewMartin:
Yeah, something like that. We could maybe do it on [this line](https://github.com/Shopify/twine/blob/master/lib/assets/javascripts/twine.coffee#L78) with something like
```coffeescript
if fn
  if type == 'bind'
    element.bindings.unshift(fn)
  else
    elements.bindings.push(fn)
```

Then there's no need to go searching for the bind each time, just keep them in the correct order.

This is assuming it's the only order dependency. Bind is also being evaluated and initializing values strictly before `define` (and eval) and that might cause problems, but only if we use bind on the same element as a define. No other bindings get evaluated during initial binding so they shouldn't cause problems (except `define-array` but that strictly needs to be run before everything else)

lib/assets/javascripts/twine.coffee
```diff
@@ -69,11 +69,16 @@
...
+    element.bindings ?= []
+    element.indexes ?= indexes
+    Array.prototype.slice.call(node.attributes).forEach (attribute) ->
```
DrewMartin: Is this faster than using a plain old for loop?

pushrax: I would expect a for loop to be fine. Alternatively `[].forEach.call` by I still don't see why.

lib/assets/javascripts/twine.coffee
pushrax: I don't see a reason to eagerly create the twine element.

pushrax:
@DrewMartin the suggested change to unshift `bind` won't work since the problem is the event handler creation order, which is done at bind-time. Instead we could keep track of a list of binding *constructors* and then defer calling them in the correct order:

```coffee
nodeBindingCtors = null # even [] is not free and most nodes have no bindings
for attrib in node.attributes
  # type = stuff
  else if fn = Twine.bindingTypes[type]
    nodeBindingCtors ?= []
    if type == 'bind' then nodeBindingCtors.unshift(fn) else nodeBindingCtors.push(fn)

if nodeBindingCtors
  for ctor in nodeBindingCtors
    # construct binding
    # element.bindings.push bla bla bla

**Comment by pushrax** [comment]
nb: reduction _to_ 70% not _by_ 70% but if you can make that happen then 😻

File: lib/assets/javascripts/twine.coffee (hunk 1 above)
Comment (pushrax) [inline lib/assets/javascripts/twine.coffee]: Probably best:

(at start)
```coffee
element = null
```

(everywhere `element` is needed)
```coffee
element ?= findOrCreateElementForNode(node)
```

justinthec:
@pushrax woops, edited 😂

I'm at the UX Summit today so I can't rebase and clean up this PR until tonight.

test/twine_test.coffee
```diff
@@ -164,6 +164,28 @@ suite "Twine", ->
...
+    test "should be the first binding to run on change event", ->
+      testView = "<input type=\"text\" bind-event-change=\"eventFunc()\" data-bind=\"val\">"
+      node = setupView(testView, context = {
+        val: 1,
+        eventFunc: () -> this.val = this.val * 2
+      })
+      node.value = 2
+      triggerEvent node, "change"
+
+      assert.equal context.val, 4
```
justinthec: This test fails when the `type == 'bind'` --> `unshift` logic (`twine.coffee:83 - 85`) is removed from the `bind` function.
```coffeescript
      if type == 'bind'
        bindingConstructors.unshift({constructor, definition})
      else
        bindingConstructors.push({constructor, definition})
```
gives: 
image

while just

```coffeescript
      bindingConstructors.push({constructor, definition})
```

gives:
image

test/twine_test.coffee
```diff
@@ -164,6 +164,28 @@ suite "Twine", ->
...
+    # test "should be the first binding to run on input event", ->
+    #   testView = "<input type=\"text\" bind-event-input=\"eventFunc()\" data-bind=\"bindFunc()\">"
+    #   node = setupView(testView, context = {
+    #     eventFunc: @spy(),
+    #     bindFunc: @spy()
+    #   })
+    #   context.eventFunc.reset()
+    #   context.bindFunc.reset()
+    #   triggerEvent node, "input"
+    #   sinon.assert.callOrder context.bindFunc, context.eventFunc
```
justinthec: I tried getting this to work but giving `data-bind` a function disabled two-way binding and therefore `bind` did not attach event listeners to the `input`, `change`, or `keyup` events.

Any ideas here? I like this test better since the intended behaviour is clearly expressed compared to the multiplication test above.

test/twine_test.coffee
```diff
@@ -164,6 +164,28 @@ suite "Twine", ->
       node = setupView(testView, key: "<script>")
       assert.equal node.innerHTML, "&lt;script&gt;"
 
+    test "should be the first binding to run on change event", ->
+      testView = "<input type=\"text\" bind-event-change=\"eventFunc()\" data-bind=\"val\">"
+      node = setupView(testView, context = {
```
lemonmade: Please pull the assignment out and above, on its own line.

test/twine_test.coffee
lemonmade: You can maybe declare `bindFunc` as a getter instead of a method, and use the spy as the getter function? You might also need to define the setter, or it might just not work, in which case I think the above test is probably fine.

lib/assets/javascripts/twine.coffee
```diff
@@ -69,13 +70,29 @@
...
+    for attribute in node.attributes
+      type = attribute.name
+      type = type.slice(5) if type.slice(0, 5) == 'data-'
+      definition = attribute.value
+
+      constructor = Twine.bindingTypes[type]
+      continue unless constructor
+
+      bindingConstructors ?= []
+      if type == 'bind'
```
lemonmade: Is this for sure the only order-dependant binding type?

justinthec:
I've rerun the trials after the new changes:

| Trial     | Before    | After    | After (new)   |
|-----------|-----------|----------|----------|
| 1         | 148.07ms  | 129.85ms | 141.06ms |
| 2         | 155.89ms  | 112.52ms | 115.91ms |
| 3         | 170.32ms  | 123.46ms | 103.59ms |
| 4         | 155.71ms  | 150.73ms | 134.22ms |
| 5         | 168.24ms  | 118.74ms | 105.80ms |
|           |           |          |          |
| Averages: | **159.646ms** | **127.06ms** | **120.116 ms** |

It looks like it got even faster 😮

lib/assets/javascripts/twine.coffee
tobi: Sorry im in mobile and can't comment on the right line somehow. Regarding line 76: slice makes an entire new string object. I recommend you test if the 5th character of he attribute is '-' before you do the slice. That's a much faster way of testing for high probability of it being a data- attribute.

lib/assets/javascripts/twine.coffee
```diff
@@ -56,6 +56,7 @@
 
   bind = (context, node, indexes, forceSaveContext) ->
     currentBindingCallbacks = []
+    element = null
```
DrewMartin: Is this actually needed? The js should have a `var element` at the top even if we don't initialize it to null

lib/assets/javascripts/twine.coffee
bouk: Or you could automatically add all the `data-` prefixed versions in the `bindingTypes` object and avoid the branching/slicing altogether 😄

lib/assets/javascripts/twine.coffee
```diff
@@ -69,13 +70,29 @@
       element = findOrCreateElementForNode(node)
       element.indexes = indexes
 
-    for type, binding of Twine.bindingTypes when definition = Twine.getAttribute(node, type)
-      element = findOrCreateElementForNode(node)
+    bindingConstructors = null
```
DrewMartin: Same with this

lib/assets/javascripts/twine.coffee
pushrax: This seems worthwhile to optimize to check each character manually

lib/assets/javascripts/twine.coffee
pushrax: I like being explicit about scope, otherwise it seems like a side-effect that the variable is being set at all

lib/assets/javascripts/twine.coffee
justinthec: This one is needed since we want to use a `?=` inside the for loop.

```
error: the variable "bindingConstructors" can't be assigned with ?= because it has not been declared before
                  bindingConstructors ?= []
```

lib/assets/javascripts/twine.coffee
```diff
@@ -69,13 +70,29 @@
...
+      type = attribute.name
+      type = type.slice(5) if type.slice(0, 5) == 'data-'
+      definition = attribute.value
+
+      constructor = Twine.bindingTypes[type]
+      continue unless constructor
+
+      bindingConstructors ?= []
+      if type == 'bind'
+        bindingConstructors.unshift({constructor, definition})
```
pushrax: Would use an array for this, they are quite a bit lighter.

lib/assets/javascripts/twine.coffee
DrewMartin: I made a quick perf test to compare some different ways. Checking each char like @justin-gomes suggested is fastest:
```javascript
indexOfPerf = function(t) {
  var start = new Date;
  for (var i = 0; i < t; i++) {
    "data-bind-event".indexOf("data-") == 0;
    "class".indexOf("data-") == 0;
  }
  console.log("indexOf took", new Date - start, "ms to run", t, "times");
}

slicePerf = function(t) {
  var start = new Date;
  for (var i = 0; i < t; i++) {
    "data-bind-event".slice(5) == "data-";
    "class".slice(5) == "data-";
  }
  console.log("slice took", new Date - start, "ms to run", t, "times");
}

charAtSlicePerf = function(t) {
  var start = new Date;
  for (var i = 0; i < t; i++) {
    "data-bind-event"[4] == "-" && "data-bind-event".slice(5) == "data-";
    "class"[4] == "-" && "class".slice(5) == "data-";
  }
  console.log("char and slice took", new Date - start, "ms to run", t, "times");
}

startsWithPerf = function(t) {
  var start = new Date;
  for (var i = 0; i < t; i++) {
    "data-bind-event".startsWith("data-");
    "class".startsWith("data-");
  }
  console.log("startsWith took", new Date - start, "ms to run", t, "times");
}

eachCharPerf = function(t) {
  var start = new Date;
  for (var i = 0; i < t; i++) {
    "data-bind-event"[0] == "d" && "data-bind-event"[1] == "a" && "data-bind-event"[2] == "t" && "data-bind-event"[3] == "a" && "data-bind-event"[4] == "-";
    "class"[0] == "d" && "class"[1] == "a" && "class"[2] == "t" && "class"[3] == "a" && "class"[4] == "-";
  }
  console.log("each char took", new Date - start, "ms to run", t, "times");
}

var times = 10000000;
indexOfPerf(times);
slicePerf(times);
charAtSlicePerf(times);
startsWithPerf(times); // not supported in IE
eachCharPerf(times);
```
```
indexOf took 1248 ms to run 10000000 times
VM4517:16 slice took 588 ms to run 10000000 times
VM4517:25 char and slice took 370 ms to run 10000000 times
VM4517:34 startsWith took 742 ms to run 10000000 times
VM4517:43 each char took 5 ms to run 10000000 times
```

lib/assets/javascripts/twine.coffee
```diff
@@ -69,13 +70,29 @@
       element = findOrCreateElementForNode(node)
       element.indexes = indexes
 
-    for type, binding of Twine.bindingTypes when definition = Twine.getAttribute(node, type)
-      element = findOrCreateElementForNode(node)
+    bindingConstructors = null
+    for attribute in node.attributes
+      type = attribute.name
+      type = type.slice(5) if type.slice(0, 5) == 'data-'
+      definition = attribute.value
```
qq99: how about moving this below `continue`?

lib/assets/javascripts/twine.coffee
```diff
@@ -69,13 +70,29 @@
...
+
+      constructor = Twine.bindingTypes[type]
```
qq99: before this, `continue unless type`?

lib/assets/javascripts/twine.coffee
justinthec: Okay I was working on @bouk 's approach by pointing 2 keys to the same function but this will add some overhead to the initial `Twine.bindingTypes` creation and the performance on checking each char looks much more promising. 

I'm going to go with this.

lib/assets/javascripts/twine.coffee
justinthec: Readability and explicitness are secondary to performance in this case 👍

lib/assets/javascripts/twine.coffee
justinthec: When would `type` ever be `undefined`? You can't have empty attribute key names afaik.

lib/assets/javascripts/twine.coffee
pushrax: I've been using this for js perf https://albertxing.com/stool/

Now that I think about it, it might be still better to put the `data-*` attribs in the map, since it would avoid allocating any substrings at all.

lib/assets/javascripts/twine.coffee
qq99: my bad, I misread the intent of what was going on here

lib/assets/javascripts/twine.coffee
```diff
@@ -265,6 +282,9 @@
   isKeypath = (value) ->
     value not in ['true', 'false', 'null', 'undefined'] && keypathRegex.test(value)
 
+  isDataAttribute = (value) ->
+    value[0] == 'd' && value[1] == 'a' && value[2] == 't' && value[3] == 'a' && value[4] == '-'
```
qq99: nitpick: stylistically, easier to read vertically if formatted like
```coffee
value[0] == 'd' &&
value[1] == 'a' &&
...
```

❤️ that this also short circuits fast too

pushrax: In C this would be written to check the first 4 bytes at once with one instruction 😛 .

justinthec:
@lemonmade 's question got collapsed
>Is this for sure the only order-dependant binding type?

From what I understand **this is the case**.

The reason is due to the fact that `bind` is the only binding type that can directly modify the `context` based on the value of the input (or any other 2-way binded element) through the use of `refreshContext`.

When `Twine.refreshImmediately()` is called afterwards (via event handlers), the value of the input gets updated based on the `context` instead. If `Twine.refreshImmediately()` were to be invoked before `bind`'s `refreshContext` then the value that the user entered into the input would be overridden by the old value in the not-yet-up-to-date `context`.

Basically we need to do:
1. `context` <-- `value`
2. `context` --> `value` and `context` --> all other binding types.

The critical part of all this is that the `context` is the first thing to be updated before all other event handlers run. All other bindings (define, eval, bind-event-*, etc) rely on the `context` to be up-to-date and that's all.

I would like @DrewMartin and/or @pushrax to confirm this though before we conclude that this is safe.
If needed I will write more tests to cover cases that I am missing.

pushrax:
Actually right now the binding order is undefined since object iteration order is not guaranteed in JS. So I suppose it's working by accident right now?

Your conclusion is correct in any case.

justinthec:
Ahh but for loops on arrays have guaranteed iteration order so this is now 100% guaranteed for us. 🎉

justinthec:
Could I get a final round of 👀 ? Thanks!

bouk:
lgtm

tobi:
do you have a final benchmark?



- tobi
CEO Shopify

qq99:
🎩 of Shopify using this worked out?

justinthec:
:tophat: is good, I'm doing a JS/Memory benchmark atm with CPU throttling to amplify the results.

lib/assets/javascripts/twine.coffee
```diff
@@ -265,6 +282,13 @@
   isKeypath = (value) ->
     value not in ['true', 'false', 'null', 'undefined'] && keypathRegex.test(value)
 
+  isDataAttribute = (value) ->
```
DrewMartin: Having a separate method for this might take away the performance benefit of doing it

DrewMartin: Just checked: Nope it doesn't, at least on chrome.

DrewMartin:
:+1: `bind` is the only one that is order dependant during events. define/define-array are order-dependant but only during initial binding and define-array is always the first thing to run. Since other bindings are only evaluated on refresh it shouldn't matter if define is ordered after them, if you were to do something like
`<div bind-event-click="myObject.onClick()" define="{myObject: new ClickHandler(this)}">`
All that to say I think this will work™️

justinthec:
Full Benchmark with CPU Throttling to Low end device (5x slowdown):
https://docs.google.com/spreadsheets/d/1_jWdLMzI8U_UzZ9goK5D3rfGw5CwjS5WD11HH_jKQcQ/edit#gid=100523775

Summary:
Average `Twine.bind` time: `587.35ms` --> `530.674ms` (~57ms reduction)
Average Heap Allocation done during `Twine.bind`: `3.24 MB` --> `2.66 MB` (0.58 MB reduction)

justinthec:
I'm going to ship this at 4pm

DrewMartin:
Nothing wrong with that! Let's wait until monday to bump the gem in shopify though ;)

justinthec:
Yeah @lemonmade has advised me to wait till monday to ship 😅

PR #82, opened 2016-07-11
Gemfile.lock

justinthec:
>Run bundle install to update Gemfile.lock
Run make .all && make .uglify to update JS

@Shopify/tnt


DrewMartin:
🚢

justinthec:
I'm a bit concerned about all the other `Gemfile.lock updates`. The last one that Max did only updated `twine-rails`. Is this cool?

DrewMartin:
considering the ruby parts of it are not used in the actual library but just for pulling in the library through rubygems (and maybe test running?) it shouldn’t be a big deal

PR #83, opened 2016-07-18
Fix define order and add tests to cover behaviour

justinthec:
#### What's the Problem?
After [Bind performance](https://github.com/Shopify/twine/pull/79) shipped last week to Shopify core, where the order of binding construction was now changed to rely on the order of attribute definition, a new edge case was found for `define` in relation to `eval` that caused a new error of which 99.6% of instances were from Firefox:

image

The scenario is when the `eval` binding is constructed before `define`. `eval` executes its definition immediately and if it depends on variables in the `define` definition then it throws an error.

#### Why does this happen?

If the attributes are arranged like so,
```html
<div data-eval="foo()" define="{foo: function() {}}"></div>
```
 then `data-eval` will execute first.

However even if the attributes are arranged the other way like this: 
```html
<div define="{foo: function() {}}" data-eval="foo()"></div>
```
then **Firefox will sort the attributes alphabetically** when generating the array for `node.attributes` and therefore `data-eval` still executes first. This does not happen on Chrome though, which is why this was not caught in testing.

#### What's the fix?

 - `define` joins the ranks of `bind` as a first order binding that needs to be constructed first before all other bindings.
 - It does not matter which of the two comes first in the list of `bindingConstructors` as long as they are both in the first two.

How are you sure that `define` does not need to come after `bind` and vice versa?

Let's lay out the dependencies and side effects of each binding **constructor**:
 - **`define`**
   - Dependencies:
     - Uses `context` to evaluate the new definition values.
   - Side Effects:
     - [Assigns the new definition values to `context`.](https://github.com/Shopify/twine/blob/fix-define-order/lib/assets/javascripts/twine.coffee#L377)
     - Does not attach listeners to any events.
 - **`bind`**
   - Dependencies:
     - Uses `context` to fetch the value of the `bind` key if it exists and is a 2-way-binding.
   - Side Effects:
     - [Assigns the key-value pair of the `bind` definition to the `context` if the key doesn't already exist in the `context` and is a 2-way-binding.](https://github.com/Shopify/twine/blob/fix-define-order/lib/assets/javascripts/twine.coffee#L334)
     - [Attaches an event listener to `'input keyup change'`.](https://github.com/Shopify/twine/blob/fix-define-order/lib/assets/javascripts/twine.coffee#L342)

**If `define` executes before `bind`**, then `bind` successfully fetches the value of its key from the `context`.

**If `bind` executes before `define`**, then `bind` sets the incorrect/empty key-value pair in the `context` at first, but when `define` executes, it overrides the same key with the correct value.

In either case, the context ends up with the correct definitions and `bind` still is the first binding to attach its listener to `'input keyup change'`. Therefore `define` and `bind` do not have order dependencies between each other.

I have added tests to preserve this non-dependency between `define` and `bind` in addition to a test ensuring `define` is run before `eval`.

**For Review:**
@GoodForOneFare @DrewMartin @Shopify/tnt

### DIFF
--- a/dist/twine.js
+++ b/dist/twine.js
@@ -99,7 +99,7 @@
           bindingConstructors = [];
         }
         definition = attribute.value;
-        if (type === 'bind') {
+        if (type === 'bind' || type === 'define') {
           bindingConstructors.unshift([constructor, definition]);
         } else {
           bindingConstructors.push([constructor, definition]);
--- a/dist/twine.js
+++ b/dist/twine.js
@@ -1,6 +1,5 @@
 (function() {
-  var indexOf = [].indexOf || function(item) { for (var i = 0, l = this.length; i < l; i++) { if (i in this && this[i] === item) return i; } return -1; },
-    slice = [].slice;
+  var slice = [].slice;
 
   (function(root, factory) {
     if (typeof root.define === 'function' && root.define.amd) {
@@ -11,7 +10,7 @@
       return root.Twine = factory();
     }
   })(this, function() {
-    var Twine, arrayPointersForNode, attribute, bind, currentBindingCallbacks, defineArray, elements, eventName, findOrCreateElementForNode, fireCustomChangeEvent, getContext, getIndexesForElement, getValue, isDataAttribute, isKeypath, j, k, keyWithArrayIndex, keypathForKey, keypathRegex, len, len1, nodeArrayIndexes, nodeCount, preventDefaultForEvent, ref, ref1, refreshElement, refreshQueued, registry, requiresRegistry, rootContext, rootNode, setValue, setupEventBinding, setupPropertyBinding, stringifyNodeAttributes, valuePropertyForNode, wrapFunctionString;
+    var Twine, arrayPointersForNode, attribute, bind, bindingOrder, currentBindingCallbacks, defineArray, elements, eventName, findOrCreateElementForNode, fireCustomChangeEvent, getContext, getIndexesForElement, getValue, isDataAttribute, isKeypath, j, k, keyWithArrayIndex, keypathForKey, keypathRegex, len, len1, nodeArrayIndexes, nodeCount, preventDefaultForEvent, ref, ref1, refreshElement, refreshQueued, registry, requiresRegistry, rootContext, rootNode, setValue, setupEventBinding, setupPropertyBinding, stringifyNodeAttributes, valuePropertyForNode, wrapFunctionString;
     Twine = {};
     Twine.shouldDiscardEvent = {};
     elements = {};
@@ -63,7 +62,7 @@
       }
     };
     bind = function(context, node, indexes, forceSaveContext) {
-      var ORDERED_BINDINGS, attribute, binding, bindingConstructors, callback, callbacks, childNode, constructor, defineArrayAttr, definition, element, j, k, key, keypath, l, len, len1, len2, len3, len4, m, n, newContextKey, newIndexes, ref, ref1, ref2, ref3, ref4, ref5, type, value;
+      var _, attribute, binding, bindingConstructors, callback, callbacks, childNode, constructor, defineArrayAttr, definition, element, j, k, key, keypath, l, len, len1, len2, len3, m, newContextKey, newIndexes, ref, ref1, ref2, ref3, ref4, type, value;
       currentBindingCallbacks = [];
       element = null;
       if (node.bindingId) {
@@ -84,7 +83,6 @@
         element = findOrCreateElementForNode(node);
         element.indexes = indexes;
       }
-      ORDERED_BINDINGS = ['define', 'bind'];
       bindingConstructors = null;
       ref = node.attributes;
       for (j = 0, len = ref.length; j < len; j++) {
@@ -98,16 +96,10 @@
           continue;
         }
         if (bindingConstructors == null) {
-          bindingConstructors = {
-            others: []
-          };
+          bindingConstructors = [];
         }
         definition = attribute.value;
-        if (indexOf.call(ORDERED_BINDINGS, type) >= 0) {
-          bindingConstructors[type] = [constructor, definition];
-        } else {
-          bindingConstructors.others.push([constructor, definition]);
-        }
+        bindingConstructors.push([type, constructor, definition]);
       }
       if (bindingConstructors) {
         if (element == null) {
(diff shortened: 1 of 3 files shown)


qq99:
Amazing write-up!

One question:

>If bind executes before define, then bind sets the incorrect/empty key-value pair in the context at first, but when define executes, it overrides the same key with the correct value.

What about this scenario:

```erb
<input data-define="{foo: new Foo()}" data-bind="foo.bar">
```

where
```coffee
class Foo
  constructor: ->
    @foo = 'test'
```

If `bind` runs before `define`, I might expect `foo.bar == 'test'` in the context, but if `define` runs first, I'd expect it in the context _and_ present on the `value` of the `input` in the DOM.  Is that the case?

tsov:
Thanks for the write-up 👍 

This feels like the order of execution should be even more explicit here, especially when some browsers automatically reorder attributes. We'd want to avoid questions like above and the order should be clear. Any thoughts?

> They may appear in any order.

- https://www.w3.org/TR/html4/intro/sgmltut.html#h-3.2.2

DrewMartin:
What if instead of unshift we guarantee order with something like
```coffeescript
ORDERED_BINDINGS = [
  'define',
  'bind'
]

bindingConstructor = {other: []}
...
if type in ORDERED_BINDINGS
  bindingConstructors[type] = [constructor, definition]
else
  bindingConstructors['other'].push([constructor, definition])
...

for type in ORDERED_BINDINGS when bindingConstructors[type]?
  [constructor, definition] = bindingConstructors[type]
  binding = constructor(node, context, definition, element)
  element.bindings.push(binding) if binding

for [constructor, definition] in bindingConstructors
  binding = constructor(node, context, definition, element)
  element.bindings.push(binding) if binding
```

Then we get guaranteed order on elements of our choice. Might be a small performance hit but as long as `ORDERED_BINDINGS` is small it should be negligible

justinthec:
@qq99 In the case of the DOM not having the correct value, `Twine.refreshImmediately` or any future call to `Twine.refresh` will take care of that.

On Shopify core, and in our tests, we call `Twine.refreshImmediately` right after `Twine.bind`.

justinthec:
@drewmartin

It will be much less performant as the list of ordered bindings grows since we need to check for each `node.attributes`, whether or not it is a member of the `ORDERED_ATTRIBUTES` list and if so, assign the definition to the right place in the list.

DrewMartin:
ORDERED_ATTRIBUTES wouldn't be all of them, it would just be the list of those where order is important that need to run first (ie define and bind, maybe eval)

qq99:
what about 3 arrays?

justinthec:
Okay after the edit, Drew's solution is pretty much what I was going for and outlines my concerns about performance.

justinthec:
Do we envision new bindings (especially ones that fundamentally need to run before other bindings) to be added to Twine?

If not, then we can keep the simpler unshift solution since `define` and `bind` are agnostic of each other.

tsov:
I'd say `eval` could potentially be on the list too, although I am not 💯  on all its use cases.

qq99:
from what I've seen, typically people will define a thing, then call a method on it using `eval` @tsov

justinthec:
@DrewMartin , I've updated the branch with your suggestion.

@qq99 @tsov `eval` will still execute its definition first even if it is not placed into ORDERED_BINDINGS since it is the only binding left that executes its definition at construction time. If we were to add another binding that executes its definition at construction time then we would need to disambiguate behaviour by moving `eval` or the other binding into ORDERD_BINDINGS.

However I feel that this may be pre-mature optimization.

TheMallen:
> On Shopify core, and in our tests, we call Twine.refreshImmediately right after Twine.bind.

Doesn't this sound like a bug more than a feature? Isn't it poor for performance that we need to do a refresh of the bindings immediately after initial bind? It's also boilerplate that is unintuitive.

TheMallen:
Also, you could do something like

```
BINDING_ORDER = {
 define: 1,
  bind: 2
}
```

Which would let you do a hash key check instead of an iterative check. I think if you used `array.sort` over the bindings with this method you could get a more scalable solution.

justinthec:
@TheMallen I think the point is that we **don't** want to scale this solution. Having bindings that are dependant on other ones unnecessarily makes Twine more complicated and also degrades the performance of `Twine.bind`. Adding `array.sort` on a 2 element array is premature optimization, and right now we're focused pretty heavily on Twine performance.

For your comment about `Twine.refreshImmediately`, I think that it is by design that `bind` and `refresh` are 2 separate actions. `bind` sets up the elements and their `refresh` callbacks so that when `refresh` is called (once at the beginning, and multiple times during runtime) the context can update the DOM values.

Also, `Twine.refreshImmediately` takes about 2% of the time of `Twine.bind` so it is both a necessary and low impact task performance wise.
image
(`Twine.refreshImmediately` is the small highlighted chunk at the end of the light blue bar taking 3.65 ms. `Twine.bind` takes 169.28 in comparison.)

TheMallen:
> Adding array.sort on a 2 element array is premature optimization, and right now we're focused pretty heavily on Twine performance.

I meant you'd sort the array of `bindingConstructors`. You wouldn't have a 2 element array to sort in the first place if you were using an object.

TheMallen:
It's fine if `refreshImmediately` is much smaller, but we should still expose a method that calls both of them if we are really doing both everywhere. `Twine.initialize` or something.

DrewMartin:
I'd prefer `eval` to go in the list as well for explicitness sake. Having its order happen to be correct just seems like it's a bug waiting to happen if new things get added.

justinthec:
@TheMallen I tried a solution using `array.sort` and passing in a `bindingOrder` comparison function:
```coffeescript
  bindingOrder = ([firstType], [secondType]) ->
    ORDERED_BINDINGS = {
      define: 1,
      bind: 2,
      eval: 3
    }
    return 1 unless ORDERED_BINDINGS[firstType]
    return -1 unless ORDERED_BINDINGS[secondType]

    ORDERED_BINDINGS[firstType] - ORDERED_BINDINGS[secondType]
```

I then conducted a performance test with CPU Throttling (5x slowdown)

| Trial     | Master    | Current Solution    |  `array.sort` Solution |
|-----------|-----------|----------|----------|
| 1         | 719.00ms  | 666.11ms | 708.83ms |
| 2         | 739.66ms  | 703.91ms | 729.67ms |
| 3         | 868.25ms  | 580.68ms | 612.42ms |
| 4         | 797.67ms  | 596.93ms | 642.33ms |
| 5         | 863.62ms  | 784.29ms | 660.84ms |
|           |           |          |          |
| Averages: | **797.64ms** | **666.384ms** | **670.818ms** |

It looks like the 2 solutions are fairly equivalent performance wise and I personally prefer the `array.sort` method since it is cleaner and reduces the amount of logic in `Twine.bind`.

justinthec:
I've added the `array.sort` solution on top of the old solution but I'll revert it if we decide to go with the old one.

GoodForOneFare:
@justinthec This looks done.  Should I be tophatting today?

justinthec:
@GoodForOneFare I'll get this merged and then tag you on the Core PR.

justinthec:
I'm going to merge this if there's no more feedback.

DrewMartin:
👍 
I'm surprised by the speed using sort. I would have expected slower, comparable to the original speed before any of these changes.

PR #85, opened 2016-08-25
Add draggable attribute binding

tsov:
Theme Editor needs `data-bind-draggable` as a binding type. Any objections?

@Shopify/tnt @owilliams 
CC @Thibaut


owilliams:
👍

lemonmade:
Seems reasonable to me 👍

PR #88, opened 2016-09-30
Keypath Function

TheMallen:
## What
This PR adds support for keypath functions. 

## How
Since twine is already checking if any bind looks like a keypath, we simply check if the value we get from a keypath is a function, and execute it if it is. There's minimal overhead here because we are already performing the keypath check on every bind.

## Why
- Evaling code is slower than just executing a function
- Evaling code makes debuggers sad and code hard to step through or put breakpoints in (nobody enjoys seeing error - source: VM231234124)
- We are already running the keypath check for simple values
- We have a lot of bindings that just look like `data-bind-event-click="foo(event)"`
- We have places where the pattern of `data-bind="foo.calculatedThing"` is approximated with eval

## Examples (trivial):
**Current**
```
<a class="next-list__item" data-bind-event-click="bulkDeleteCollectionsModal.show()">Delete selected collections</a>
```
**Keypath version**
```
<a class="next-list__item" data-bind-event-click="bulkDeleteCollectionsModal.show">Delete selected collections</a>
```

**Current**
```
'data-bind-event-click' => 'nextNavigation.goTo(event)',
```
**Keypath version**
```
'data-bind-event-click' => 'nextNavigation.goTo',
```

## Comments pls
FED folks who use lots of twine:
@WillsonSmith 
@dan-menard 
Reviewer folks
@Shopify/tnt 
@pushrax 
@qq99

### DIFF
--- a/dist/twine.js
+++ b/dist/twine.js
@@ -367,21 +367,21 @@
       var e, error, keypath;
       if (isKeypath(code) && (keypath = keypathForKey(node, code))) {
         if (keypath[0] === '$root') {
-          return function($context, $root) {
+          return function($context, $root, arrayIndexes, event) {
             var value;
             value = getValue($root, keypath);
             if (typeof value === 'function') {
-              return value();
+              return value(node, event);
             } else {
               return value;
             }
           };
         } else {
-          return function($context, $root) {
+          return function($context, $root, arrayIndexes, event) {
             var value;
             value = getValue($context, keypath);
             if (typeof value === 'function') {
-              return value();
+              return value(node, event);
             } else {
               return value;
             }
@@ -622,7 +622,7 @@
       return Twine.bindingTypes["bind-event-" + eventName] = function(node, context, definition) {
         var onEventHandler;
         onEventHandler = function(event, data) {
-          var base, discardEvent;
+          var base, discardEvent, fn;
           discardEvent = typeof (base = Twine.shouldDiscardEvent)[eventName] === "function" ? base[eventName](event) : void 0;
           if (discardEvent || preventDefaultForEvent(event)) {
             event.preventDefault();
@@ -630,7 +630,8 @@
           if (discardEvent) {
             return;
           }
-          wrapFunctionString(definition, '$context,$root,$arrayPointers,event,data', node).call(node, context, rootContext, arrayPointersForNode(node, context), event, data);
+          fn = wrapFunctionString(definition, '$context,$root,$arrayPointers,event,data', node);
+          fn.call(node, context, rootContext, arrayPointersForNode(node, context), event, data);
           return Twine.refreshImmediately();
         };
         $(node).on(eventName, onEventHandler);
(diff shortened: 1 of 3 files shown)


qq99:
I think I'm ambivalent to this kind of change

I don't know that we should encourage/support 2 different ways of function calling, but that's just my opinion.

The better stacktraces is appealing though!

DrewMartin:
> Evaling code is slower than just executing a function

What kind of speed up would this give us?

dan-menard:
I'm curious about the speedup as well; my gut says it might actually be pretty significant for initial page load.

One neat opportunity for checking this: Right now we measure a whole bunch of metrics for initial page load on Home, which is a fairly heavy Twine user. If we were to merge this and rewire Home's Twine calls with the new syntax, we'd get a nice before/after picture on our dashboards and would get a good read on the size of the perf gain.

I'm happy to do the rewiring myself, and probably have time to do that this week, if that sways the argument at all :)

pushrax:
I like this idea, but I am strongly against changing the syntax. If we implement a non-eval path, IMO it should based on parsing existing syntax, like Angular does. It's not simple to implement a general parser ([example from Angular](https://github.com/angular/angular.js/blob/823295fee058f169523e84f94be13962a9513d3d/src/ng/parse.js)) but making one that works for simple function calls would be straightforward.

TheMallen:
We could do a simple parser, but it would be a tough balance between:
- Not making it so complex that it bloats the library (angular's is 2000 lines)
- Falling back to eval at the right times (ie. not breaking any of our existing bindings)
- Still avoiding eval a meaningful amount of the time

We could do something super simple like:
- check if there's a `(*)` at the end of the string, if so pull it out, they're args
- do the keypath thing
- if it's a keypath and we have an arg expression we parse the args (I guess just exploding it on comma and doing keypath stuff to each segment, or see if they match specific literals like `event`)

The naive solution is still pretty brittle though, and it's easy to think of things that are intuitive that break it (`data-bind-event="foo.someHandler(foo.someUtilityFunction(lol, broken))"`), and once you go down the path of making it robust you rapidly start to approach proper parser land...

TheMallen:
For an idea of eval vs no eval vs new function, including cases where we have an eval in scope but don't execute it, I created a (contrived) benchmark.

http://codepen.io/themallen/pen/pEdgYv?editors=1010

image

TheMallen:
So basically on chrome eval and new function are orders of magnitude slower than non-eval. At least in contrived cases.

TheMallen:
Running the codepen cases one by one a few times gives
image

pushrax:
I don't understand the significance of this benchmark – `noEval` seems to do nothing at all related to the test? Why do all the tests querySelector a div?

TheMallen:
It's literally just showing the effect of minimum effect of eval on the JIT. The queryselector is just arbitrary.

TheMallen:
Closing this for now, might take a crack at a simple parsing version later.

Issue #50, opened 2015-12-08
Remove Twine's dependency on `jQuery.toggleClass`


tylerball:
I'd like to be able to modify classes on an `<svg>`, which jquery doesn't support.

PR #89, opened 2016-10-31
Remove jQuery.toggleClass

asmockler:
Addresses #50. Should be noted that `classList.toggle` still doesn't work on SVGs in IE11. Everything else should be good according to [caniuse](http://caniuse.com/#search=classlist).

Very small change but thought I would go ahead and do it.

cc @Shopify/tnt

### DIFF
--- a/dist/twine.js
+++ b/dist/twine.js
@@ -523,7 +523,9 @@
               return;
             }
             if (lastValue = newValue) {
-              return node.classList.toggle('hide');
+              return node.classList.add('hide');
+            } else {
+              return node.classList.remove('hide');
             }
           }
         };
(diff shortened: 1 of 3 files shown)


pushrax:
The previous code would add the class if the value is truthy and remove the class if falsy. The new code does something different.

pushrax:
A test should really be failing here :s

TheMallen:
`node.classList.toggle` also takes a second parameter similar to jquery's, we should just use that.

asmockler:
IE11 [doesn't support](http://caniuse.com/#search=classlist) the second parameter for `node.classList.toggle` so we will just have to wrap it in an `if` statement. Planning on doing it soon - just keep getting distracted.

PR #91, opened 2016-11-09
Add callback to refresh()

pushrax:
As discussed. This lets you do something after the current refresh is complete.

@jonathaneckmier @TheMallen @Shopify/tnt

### DIFF
--- a/dist/twine.js
+++ b/dist/twine.js
@@ -10,7 +10,7 @@
       return root.Twine = factory();
     }
   })(this, function() {
-    var Twine, arrayPointersForNode, attribute, bind, bindingOrder, currentBindingCallbacks, defineArray, elements, eventName, findOrCreateElementForNode, fireCustomChangeEvent, getContext, getIndexesForElement, getValue, isDataAttribute, isKeypath, j, k, keyWithArrayIndex, keypathForKey, keypathRegex, len, len1, nodeArrayIndexes, nodeCount, preventDefaultForEvent, ref, ref1, refreshElement, refreshQueued, registry, requiresRegistry, rootContext, rootNode, setValue, setupEventBinding, setupPropertyBinding, stringifyNodeAttributes, valuePropertyForNode, wrapFunctionString;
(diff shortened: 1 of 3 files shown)


pushrax:
Good call, there was one...

lib/assets/javascripts/twine.coffee
```diff
@@ -124,7 +125,9 @@
     elements[node.bindingId] ?= {}
 
   # Queues a refresh of the DOM, batching up calls for the current synchronous block.
-  Twine.refresh = ->
+  # The callback will be called once when the refresh has completed.
+  Twine.refresh = (callback) ->
+    refreshCallbacks.push(callback) if callback
     return if refreshQueued
     refreshQueued = true
     setTimeout(Twine.refreshImmediately, 0)
```
tsov: Just wondering, could we do `Twine.refreshImmediately.bind(Twine, callback)` here and `Twine.refreshImmediately = (callback) ->` in https://github.com/Shopify/twine/pull/91/files#diff-5c7054f0eed5458018a14692e6e48d93R139?

That way we wouldn't need the `refreshCallbacks` variable and both `refresh` and `refreshImmediately` would have the same api interface

tsov: Edit: never mind me, `refreshImmediately` is synchronous anyway

gif-keyboard-14694901250535498836

pushrax: `refreshImmediately` is synchronous so it's a bit of an antipattern to give it a callback

We need the global list of callbacks since a call to `Twine.refresh` will not execute `refreshImmediately` if there is a pending refresh already.

DrewMartin:
A test case that could be added:
if a callback triggers a `Twine.refresh` with a callback, that callback won't be called until the next execution of `refreshImmediately`

DrewMartin:
👍

dist/twine.js
```diff
@@ -181,12 +185,18 @@
...
       refreshQueued = false;
       for (key in elements) {
         element = elements[key];
         refreshElement(element);
       }
+      callbacks = refreshCallbacks;
+      refreshCallbacks = [];
+      for (j = 0, len = callbacks.length; j < len; j++) {
+        cb = callbacks[j];
+        cb();
```
TheMallen: So, we're never removing them from the array. Does that always make sense?

I think there are places in our code right now using `refreshImmediately` and then doing something right afterwards expecting it to be done, but they don't necessarily want the code to be run after *every* refresh.

Maybe we should view a passed callback as wanting to execute after *this* refresh, and then provide a `twine:after-refresh` event to bind to for cases where you might want to have something happen after *every* refresh?

PR #93, opened 2017-02-09
Replace $ with jQuery

dan-menard:
This will allow apps that already have something else mapped to `$` to use Twine without it conflicting due to Twine assuming jQuery lives on `window.$`

### DIFF
--- a/lib/assets/javascripts/twine.coffee
+++ b/lib/assets/javascripts/twine.coffee
@@ -354,9 +354,9 @@
           return if getValue(context, keypath) == this[valueProp]
           refreshContext()
           Twine.refreshImmediately()
-        $(node).on 'input keyup change', changeHandler
+        jQuery(node).on 'input keyup change', changeHandler
         teardown = ->
-          $(node).off 'input keyup change', changeHandler
+          jQuery(node).off 'input keyup change', changeHandler
 
       {refresh, teardown}
 
@@ -366,12 +366,12 @@
       return refresh: ->
         newValue = !fn.call(node, context, rootContext, arrayPointersForNode(node, context))
         return if newValue == lastValue
-        $(node).toggleClass('hide', lastValue = newValue)
+        jQuery(node).toggleClass('hide', lastValue = newValue)
 
     'bind-class': (node, context, definition) ->
       fn = wrapFunctionString(definition, '$context,$root,$arrayPointers', node)
       lastValues = {}
-      $node = $(node)
+      $node = jQuery(node)
       return refresh: ->
         newValues = fn.call(node, context, rootContext, arrayPointersForNode(node, context))
         additions = []
@@ -396,7 +396,7 @@
       return refresh: ->
         newValue = fn.call(node, context, rootContext, arrayPointersForNode(node, context))
         for key, value of newValue when lastValue[key] != value
-          $(node).attr(key, value || null)
+          jQuery(node).attr(key, value || null)
         lastValue = newValue
 
     define: (node, context, definition) ->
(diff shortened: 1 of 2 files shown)


TheMallen:
I think we should just use injection instead of relying on a different global. We already have a UMD wrapper so it's pretty easy to do.

I'm thinking something along the lines of:

```
((root, factory) ->
  if typeof root.define == 'function' && root.define.amd
    root.define(['jquery'], factory)
  else if typeof module == 'object' && module.exports
    jQuery = require('jquery')
    module.exports = factory(jQuery)
  else
    root.Twine = factory(jQuery)
)(this, ($) ->
```

TheMallen:
That said, this will totally work fine, and I don't feel *that* strongly about it.

qq99:
I like @TheMallen's suggestion, though I don't feel super strongly either way

dan-menard:
Updated! Now with 100% more @TheMallen influence.

I tested this in a large module-based app, so I know the `module.exports` case works. Is there an easy way to make sure this won't break for anyone running AMD? Or can someone familiar with AMD confirm that what I have there makes sense?

TheMallen:
It would work with amd provided the consumer has jquery available as an amd module already.

dan-menard:
So CI wasn't passing, because it couldn't find `jquery` in our dependency tree, which is legit because it wasn't there. That seems wrong to me, Twine absolutely depends on jQuery. So I added jQuery 2.2.0 as a dependency in our `package.json`.

@pushrax @qq99 that's cool, right? Any concerns?

package.json
```diff
@@ -27,6 +27,9 @@
     "url": "https://github.com/Shopify/twine/issues"
   },
   "homepage": "https://github.com/Shopify/twine",
+  "dependencies": {
+    "jquery": "2.2.0"
```
GoodForOneFare: This is going to add a `jquery@2.2.0` dependency to anything consuming Twine.  Chances are that npm will resolve to the user's preferred version, but this still seems odd to me.

If we're not directly including it as part of Twine, maybe it'd be better as a `peerDependency`, with a `^2.2.0` to make it less specific?  

cc @lemonmade because I know you don't like peers :P

lemonmade:
I'd probably stick to a dependency (it is a dependency after all) but a much looser one (as loose as we can) so that we don't risk situations where we end up having two versions of jQuery (application's + Twine's more specific one, if the version ranges are incompatible or they are using older npm).

pushrax:
Twine should work with pretty much any jQuery right?

dan-menard:
I'll be honest, I don't have a good grasp on the finer points of NPM-like dependency management so I just yolo'd the version we have in Shopify core. I agree mostly with Chris, a normal dependency seems right in this case, so it's just a matter of picking the right major version.

Looking [here](https://w3techs.com/technologies/details/js-jquery/all/all) `jquery 1.*` is _vastly_ more popular than `jquery 2.*`, and if we [dig in further](https://w3techs.com/technologies/details/js-jquery/1/all), it looks like almost everyone is using at least `jquery 1.0`.

How do we feel about `>= 1.*`?

dan-menard:
Reverted the PR to the simpler `$` -> `jQuery` conversion.

I don't know what happened to the tests, but loading jQuery as a proper dependency seems to cause many of them to fail. As far as I could tell, nothing was wrong with the Twine build (it worked just fine) and after a day of digging, I still have no idea what's causing the failures.

I know this is a lesser solution, but it still fixes a major problem in Twine (Twine crashes internally if some other lib, like prototype, is mapped on `window.$` instead of jQ). Hopefully we can ship this as a workaround for that until we can remove jQuery entirely.

@TheMallen @pushrax good to go? This is the same change I pitched when I first opened the PR.

pushrax:
LGTM

PR #95, opened 2018-02-15
Add UMD support for jQuery component injection

mattdrose:
This injects `jquery` in the module definition.

Right now Twine is relying on a global definition. Marketing is converting Shopify's brochure to es6 components (signup uses Twine), so `jquery` will only be available as a defined component.

This also adds jQuery as a proper dependency, and removes the vendor.


lib/assets/javascripts/twine.coffee
```diff
@@ -1,11 +1,15 @@
 ((root, factory) ->
   if typeof root.define == 'function' && root.define.amd
-    root.define([], factory)
+    root.define(['jquery'], factory)
   else if typeof module == 'object' && module.exports
-    module.exports = factory()
+    jQuery = if typeof window != 'undefined'
+    then require('jquery')
+    else require('jquery')(root)
```
tsov: I am assuming this is for SSR? ^

mattdrose: Not sure what SSR stands for 😢?

When the `window` object isn't present, [jQuery returns a factory rather than the direct module](https://github.com/jquery/jquery/blob/master/src/wrapper.js#L21-L35).

tsov: Oh sorry, I meant Server Side Rendering. 👍

mattdrose: [Here's an example of the same tactic used in a UMD template](https://github.com/umdjs/umd/blob/master/templates/jqueryPlugin.js#L9-L24)

pushrax:
Thanks for this!

PR #96, opened 2018-02-15
Bump to 1.2.1

mattdrose:
Bump includes changes from https://github.com/Shopify/twine/pull/95


dev.yml
```diff
@@ -2,7 +2,7 @@ name: twine
 
 up:
   - node:
-      version: v5.7.0
+      version: v6.11.1
```
mattdrose: `dev` doesn't support v5 due to security vulnerabilities.

mattdrose:
@tsov this shouldn't break any dependants. It'll fallback to global definition of jQuery if amd or commonjs don't exist. If it is a breaking change, we should bump to 2.0.0 (according to semver). I bumped a patch thinking that this is fixing expected behaviour with UMD.

I can 🎩 if you have any apps in mind.

PR #99, opened 2018-10-22
Don’t iterate over dynamically added nodes for bind/unbind

lemonmade:
This PR prevents iterating over dynamically inserted nodes during `Twine.bind`/ `Twine.unbind` by creating a locked array of children before starting iteration. This is in order to address an issue in our main app (https://github.com/Shopify/web/issues/7333, for those who have access), but more generally I feel this is a bug of the current implementation, not a feature. As such, I'd propose this as a patch release.

### DIFF
--- a/lib/assets/javascripts/twine.coffee
+++ b/lib/assets/javascripts/twine.coffee
@@ -110,12 +110,9 @@
 
     callbacks = currentBindingCallbacks
 
-    # IE and Safari don't support node.children for DocumentFragment and SVGElement nodes.
-    # If the element supports children we continue to traverse the children, otherwise
-    # we stop traversing that subtree.
-    # https://developer.mozilla.org/en-US/docs/Web/API/ParentNode.children
-    # As a result, Twine are unsupported within DocumentFragment and SVGElement nodes.
-    bind(context, childNode, if newContextKey? then null else indexes) for childNode in (node.children || [])
+    # IE and Safari don't support node.children for DocumentFragment or SVGElement,
+    # See explanation in childrenForNode()
+    bind(context, childNode, if newContextKey? then null else indexes) for childNode in childrenForNode(node)
     Twine.count = nodeCount
 
     for callback in callbacks || []
@@ -123,6 +120,18 @@
     currentBindingCallbacks = null
 
     Twine
+
+  # IE and Safari don't support node.children for DocumentFragment and SVGElement nodes.
+  # If the element supports children we continue to traverse the children, otherwise
+  # we stop traversing that subtree.
+  # https://developer.mozilla.org/en-US/docs/Web/API/ParentNode.children
+  # As a result, Twine are unsupported within DocumentFragment and SVGElement nodes.
+  #
+  # We also prevent nodes from being iterated over more than once by locking
+  # the children, which prevents nodes that are dynamically inserted as siblings
+  # from causing double/ missed binds and unbinds.
+  childrenForNode = (node) ->
+    if node.children then Array.prototype.slice.call(node.children, 0) else []
 
   findOrCreateElementForNode = (node) ->
     node.bindingId ?= ++nodeCount
@@ -169,10 +178,9 @@
       delete elements[id]
       delete node.bindingId
 
-
     # IE and Safari don't support node.children for DocumentFragment or SVGElement,
-    # See explaination in bind()
-    Twine.unbind(childNode) for childNode in (node.children || [])
+    # See explanation in childrenForNode()
+    Twine.unbind(childNode) for childNode in childrenForNode(node)
     this
 
   # Returns the binding context for a node by looking up the tree.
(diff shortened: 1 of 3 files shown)


lib/assets/javascripts/twine.coffee
```diff
@@ -124,6 +121,17 @@
...
     Twine
 
+  # IE and Safari don't support node.children for DocumentFragment and SVGElement nodes.
+  # If the element supports children we continue to traverse the children, otherwise
+  # we stop traversing that subtree.
+  # https://developer.mozilla.org/en-US/docs/Web/API/ParentNode.children
+  # As a result, Twine are unsupported within DocumentFragment and SVGElement nodes.
+  #
+  # We also prevent nodes from being iterated over more than once by locking
+  # the children, which prevents
```
qq99: sentence stops abruptly here

lib/assets/javascripts/twine.coffee
```diff
@@ -124,6 +121,18 @@
...
+  # If the element supports children we continue to traverse the children, otherwise
+  # we stop traversing that subtree.
+  # https://developer.mozilla.org/en-US/docs/Web/API/ParentNode.children
+  # As a result, Twine are unsupported within DocumentFragment and SVGElement nodes.
+  #
+  # We also prevent nodes from being iterated over more than once by locking
+  # the children, which prevents nodes that are dynamically inserted as siblings
+  # from causing double/ missed binds and unbinds.
+  childrenForNode = (node) ->
+    if node.children then Array.prototype.slice.call(node.children, 0) else []
```
nickjs: `Array::slice.call(node.children)`

lib/assets/javascripts/twine.coffee
```diff
@@ -124,6 +121,18 @@
...
 
+  # IE and Safari don't support node.children for DocumentFragment and SVGElement nodes.
+  # If the element supports children we continue to traverse the children, otherwise
+  # we stop traversing that subtree.
+  # https://developer.mozilla.org/en-US/docs/Web/API/ParentNode.children
+  # As a result, Twine are unsupported within DocumentFragment and SVGElement nodes.
+  #
+  # We also prevent nodes from being iterated over more than once by locking
+  # the children, which prevents nodes that are dynamically inserted as siblings
+  # from causing double/ missed binds and unbinds.
```
nickjs: Do you know if any functionality is depending on this? If we're currently double binding, is it possible people have put workarounds into their code to, for example, remove one bind?

lib/assets/javascripts/twine.coffee
```diff
@@ -124,6 +121,18 @@
 
     Twine
 
+  # IE and Safari don't support node.children for DocumentFragment and SVGElement nodes.
+  # If the element supports children we continue to traverse the children, otherwise
+  # we stop traversing that subtree.
+  # https://developer.mozilla.org/en-US/docs/Web/API/ParentNode.children
+  # As a result, Twine are unsupported within DocumentFragment and SVGElement nodes.
+  #
+  # We also prevent nodes from being iterated over more than once by locking
```
GoodForOneFare: "locking" is an odd term that makes it sound like something special is happening .  I'd expect something like, `A static list of current DOM nodes is iterated over, preventing immediately inserted/ deleted siblings from causing double/ missed binds and unbinds`.

lib/assets/javascripts/twine.coffee
```diff
@@ -110,12 +110,9 @@
...
 
-    # IE and Safari don't support node.children for DocumentFragment and SVGElement nodes.
-    # If the element supports children we continue to traverse the children, otherwise
-    # we stop traversing that subtree.
-    # https://developer.mozilla.org/en-US/docs/Web/API/ParentNode.children
-    # As a result, Twine are unsupported within DocumentFragment and SVGElement nodes.
-    bind(context, childNode, if newContextKey? then null else indexes) for childNode in (node.children || [])
+    # IE and Safari don't support node.children for DocumentFragment or SVGElement,
+    # See explanation in childrenForNode()
+    bind(context, childNode, if newContextKey? then null else indexes) for childNode in childrenForNode(node)
```
GoodForOneFare: I guess we could compare the before/after list and warn about it in dev mode?  Doesn't seem worth it to me, but I don't have to troubleshoot Twine issues 😛

lib/assets/javascripts/twine.coffee
pushrax: "caching a copy" would be clear to me.

qq99:
As @lemonmade described it to me, it would be surprising if someone were relying on this behaviour -- if anything I figure other people were also suffering from this (perhaps without realizing it)

lemonmade:
I couldn't see a meaningful perf impact when I tried this out locally, I think it basically pales in comparison to how slow the DOM operations are /shrug

lib/assets/javascripts/twine.coffee
lemonmade: I think it should be OK, we'd still have to tophat everything to see those messages

lib/assets/javascripts/twine.coffee
lemonmade: Nothing during my general tophatting appeared broken, I can't guarantee that it does not but I think it's just as likely something is silently broken by this (example: you *remove* a sibling element when during a bind on the second last element in the list, now you have already iterated over the current size of the list so the binding stops, even though the last item was never traversed)

PR #102, opened 2021-02-18
Remove jQuery

brendo:
_Firsty, many thanks to those that review this PR on a fairly antiquated codebase 🙏_

This PR removes jQuery from Twine. It's part of a greater effort to remove jQuery from shopify.com. After researching our dependency chain we realised that until jQuery is removed from Twine we wouldn't be able to achieve that goal. 

This PR contains a few "random" commits needed to work on the project locally and in CI (ea80b36 and 3751e55).

Otherwise, cc22591 migrates the event handlers from `on/off` and moves to native event handling. This required the most amount of changes so was isolated from the remaining migration in 929acdc.

As far as I can tell this change is compatible with earlier Twine releases with two exceptions:

- Using `jQuery.trigger(eventName, data)` will no longer work. This was used in Twine tests but I haven't seen it in the wild yet.
- `data-bind-attribute` will no longer process values that are functions. This behaviour seemed to be inadvertently possible because this logic will fall to jQuery's `attr` method which would handle executing functions before setting the resulting value. A Grok search for `data-bind-attribute` across Shopify's codebases didn't show any usage for this behaviour but I wouldn't say that was super exhaustive.

Given those two "regressions", I'd probably recommend that this should be released as Twine v2.


brendo:
I have tophatted this on Brochure2 by checking that 3F still works as expected (this is the only place in Brochure we use Twine) 🎉

test/twine_test.coffee
```diff
@@ -644,43 +642,33 @@ suite "Twine", ->
   suite "data-allow-default", ->
     test "should prevent default action for an anchor tag", ->
       testView = "<a data-bind-event-click=\"fn()\"></a>"
-      event = {type: 'click', preventDefault: @spy()}
       node = setupView(testView, fn: ->)
 
-      jQuery(node).trigger(event)
-      assert.isTrue event.preventDefault.calledOnce
+      assert.isFalse triggerEvent(node, "click")
```
brendo, line 647: This took a while to figure out how to resolve but after some searching the [MDN `dispatchEvent` docs](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent) revealed that the result of `dispatchEvent` will be:

> The return value is false if event is cancelable and at least one of the event handlers which received event called Event.preventDefault(). Otherwise it returns true.

This works perfectly for our scenario as the previous tests were asserting the dummy "event" object's spy was called.

lib/assets/javascripts/twine.coffee
```diff
@@ -462,19 +457,19 @@
 
   setupEventBinding = (eventName) ->
     Twine.bindingTypes["bind-event-#{eventName}"] = (node, context, definition) ->
-      onEventHandler = (event, data) ->
+      onEventHandler = (event) ->
```
mattdrose, line 460: I think removing `data` on 472 could could potentially break consumers if they rely on it. [You can access data on custom events via `.detail`.](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/detail)

```suggestion
      onEventHandler = (event, data = event.detail) ->
```

mattdrose (commented):
Long live twine. I guess this is a major bump? Since it drops old ie support.

qq99:
>Using jQuery.trigger(eventName, data) will no longer work

I think there must be spots within admin doing things like `$('.some-selector').trigger('click')` or `$someNode.trigger(...)`.  It's been a heck of a long time, but I believe that's roughly equivalent to `jQuery.trigger(...)`.  I'd argue that we should not be bumping Twine within admin at this point, so while this is a breaking change of sorts

>A Grok search for data-bind-attribute across Shopify's codebases didn't show any usage for this behaviour but I wouldn't say that was super exhaustive

You probably also need to search for `bind-attribute` as well (IIRC, at one point I believe we didn't have namespace our attributes with `data-*`)

qq99:
I think we will also want to bump the Twine version, maybe using a major version in this case

lib/assets/javascripts/twine.coffee
```diff
@@ -378,28 +375,23 @@
...
           newValue = newValues[key] = !!newValues[key]
-          currValue = lastValues[key] ? $node.hasClass(key)
+          currValue = lastValues[key] ? node.classList.contains(key)
           if currValue != newValue
             if newValue
-              additions.push(key)
+              node.classList.add(key)
             else
-              removals.push(key)
+              node.classList.remove(key)
```
qq99, line 393: ✅ https://caniuse.com/classlist looks pretty good to me

brendo:
> I'd argue that we should not be bumping Twine within admin at this point, so while this is a breaking change of sorts, we don't need to break admin.

Definitely agree.

I'll push another commit which brings back `data` using `event.detail`. This won't prevent the regression as users would still have to modify their application to start dispatching custom events but at least there is some sort of upgrade path if folks are in that position.

I'll then push a "final" commit releasing this branch as v2.0.

lib/assets/javascripts/twine.coffee
brendo: Done. I couldn't accept this suggestion as I had to make other changes in this method to make things tick 👍

brendo:
I've pushed 9e91930 which bumps to v2.0.0 as per the [Releasing notes](https://github.com/Shopify/twine#releasing).

Steps 4 & 5 look like they need to be done once the branch is in `master` (and I'm a bit unclear if I'll actually have permissions to run them locally 😅)

qq99:
>Steps 4 & 5 look like they need to be done once the branch is in master

IIRC yes

brendo:
Great, well let me know how you want to proceed :) From the looks of things @lemonmade was the last to push a release (back in 2018 🙈)

qq99:
https://github.com/Shopify/twine/releases as well

lib/assets/javascripts/twine.coffee
```diff
@@ -462,7 +457,7 @@
 
   setupEventBinding = (eventName) ->
     Twine.bindingTypes["bind-event-#{eventName}"] = (node, context, definition) ->
-      onEventHandler = (event, data) ->
+      onEventHandler = (event, data = event.detail) ->
```
lemonmade, line 460: Is this something that jQuery does?

package.json
```diff
@@ -1,6 +1,6 @@
 {
```
lemonmade, line 1: Any particular reason you chose to stick to `npm` instead of using `yarn` to install the dependencies (based on the lockfile above)?

lemonmade (approved):
Thanks for the fun refresher on jQuery and CoffeeScript 😄 

Changes LGTM, and I agree with 2.0, since it's hard to guess at whether people depend on the transitive jQuery dependency here. If you need help with the release, let me know! Thanks for digging through these historical artifacts and doing the dirty work in the name of performance :)

lib/assets/javascripts/twine.coffee
qq99: I think I found it in https://api.jquery.com/on/
Screen Shot 2021-02-22 at 10 52 57 AM

We use this in admin via `$someEl.trigger('click', someObj)`

qq99 (approved):

test/twine_test.coffee
```diff
@@ -350,16 +348,17 @@ suite "Twine", ->
       node = setupView(testView, context = fn: @spy())
       Twine.shouldDiscardEvent.click = -> true
 
-      jQuery(node).click()
+      triggerEvent node, "click"
```
one20, line 351: There's instances of `triggerEvent(node, "click")` here too (with parenthesis), should they be consistent?

one20 (approved):
I have very little context on this package, i forgot it even existed. To the best of my knowledge things look good to me based on the changes on this PR.

lib/assets/javascripts/twine.coffee
brendo: Yep, that's it @qq99. I originally removed the data functionality but @mattdrose had the idea to make use of the native [`detail` property](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/detail) of `CustomEvent`. This means that folks _could_ upgrade to something similar [as shown in this test](https://github.com/Shopify/twine/blob/5f28a75694c354aa9127868faf797026efa24523/test/twine_test.coffee#L355-L364).

package.json
brendo: I was trying to rock the boat at least as possible 😅 I assumed (perhaps wrongly) that Yarn wouldn't be available in Travis and noticed that the Contributing guidelines made reference to `npm install` so stuck with NPM.

